PR #5, opened 2016-03-31
Changes for v2

lox:
This PR moves to using the v2 of the Buildkite API as well as moving from projects to pipelines.


lox:
@toolmantim @wolfeidau is this a good point to move this over to the buildkite org?

toolmantim:
Woo! Yeah I can do this if @wolfeidau adds me as an admin

wolfeidau:
@toolmantim I have added you but I think this is just for push access.

wolfeidau:
@toolmantim I can transfer this to you and then you can push it to buildkite if you like.

toolmantim:
@wolfeidau I think you might need to do that

toolmantim:
@lox merge when ready!

lox:
Done, tag away good sir.

wolfeidau:
OK updated the version number and tagged https://github.com/buildkite/go-buildkite/releases/tag/v2.0.0 

Just need to tweak the readme.

PR #6, opened 2016-03-31
Added BadgeURL property to Pipeline

tobyjoe:
Small change to add the BadgeURL property that @keithpitt added on 3/28

:hammer:


keithpitt:
:+1:

PR #7, opened 2016-03-31
Removed the deprecated/dead featured_build pipeline property

tobyjoe:
From what I can tell, the ```featured_build``` attribute from the pipeline API is no longer with us, so I removed it off the v2.0.0 tag.

:skull:

### DIFF
--- a/buildkite/pipelines.go
+++ b/buildkite/pipelines.go
@@ -35,9 +35,6 @@ type Pipeline struct {
 	// the provider of sources
 	Provider *Provider `json:"provider,omitempty"`
 
-	// build featured when you view the pipeline
-	FeaturedBuild *Build `json:"featured_build,omitempty"`
-
 	// build steps
 	Steps []*Step `json:"steps,omitempty"`
 }


wolfeidau:
@keithpitt Is this the case, if so I will merge.

Cheers

keithpitt:
That's right! We realised that when you're triggering builds with a ton of people on the same pipeline, `featured_build` ends up changing very often, and it's pretty useless, so we've dropped it from the API. In the next iteration of our pipelines list page, we'll be dropping it from there as well.

So yeah! Feel free to kill it :skull:

sr:
@keithpitt It would be nice to allow clients to retrieve this information in a single API request —finding the status of a project is a pretty common use case. That is, the status of its latest "master" (or equivalent, whatever is set as the "default branch" on GitHub) branch build.

I know performance of the query required to get this data sucks... Perhaps still could be provided as a separate endpoint that would return the statuses of the most recent builds of the given branch for all the pipelines.  Example:

```
GET /v2/organizations/{org.slug}/pipelines_statuses?branch={branch_name}
```

Anyway, just a though. Cheers!

keithpitt:
@sr totally, that's where our GraphQL API comes in :) You could do a query like this to get the last 5 builds for the `master` branch in the `buildkite/frontend` pipeline

```graphql
query {
  pipeline(slug: "buildkite/frontend") {
    builds(branch: "master", first: 5) {
      edges {
        node {
          state
        }
      }
    }
  }
}
```

Which will return:

```json
{
  "data": {
    "pipeline": {
      "builds": {
        "edges": [
          {
            "node": {
              "number": 388,
              "message": "Bumped emojis",
              "state": "passed"
            }
          },
          {
            "node": {
              "number": 387,
              "message": "Pipeline styles",
              "state": "failed"
            }
          },
          {
            "node": {
              "number": 386,
              "message": "Merge branch 'master' into organization-show",
              "state": "failed"
            }
          },
          {
            "node": {
              "number": 385,
              "message": "The chevron-right icon should use currentColor as the stroke color",
              "state": "passed"
            }
          },
          {
            "node": {
              "number": 384,
              "message": "New chevron-right (again)",
              "state": "passed"
            }
          }
        ]
      }
    }
  }
}
```

PR #9, opened 2016-03-31
Changed the wolfeidau refs to buildkite

tobyjoe:
The ```makefile``` and ```main.go``` files had refs to ```github.com/wolfeidau``` still, so I changed them to ```github.com/buildkite```


wolfeidau:
Yeah I need to port this to my new docker based build strategy.

Basically pull down the golang docker container, rejig some directories and run tests.

The infra in the `Makefile` was to enable build in place and it is a bit crappy.

PR #10, opened 2016-04-26
Add Artifacts Service

pquerna:
- Adds `ArtifactsService` struct to Client.
- Changes behavior of the Auth RoundTrippers, to only pass in Authorization headers if the `.Host` matches the original API host.  This is because we were unconditionally adding the Authorization Headers on redirects to an s3 URL.  This has two bad things:
  * Leaking auth headers to Amazon S3
  * This would cause Amazon to 400, because the redirect URL includes Amz params:
```xml
<Error>
<Code>InvalidArgument</Code>
<Message>Only one auth mechanism allowed;
only the X-Amz-Algorithm query parameter,
Signature query string parameter or the Authorization header should be specified</Message>
```

So, I did a horrible hack on the auth round trippers, open to another idea, but basically anything I could think of would break the existing API.  I did try using the `CheckRedirect` on the http.Client, but it is invoked *before* the Transport, so deleting the Authorization header has no effect, as the transport would just add the Authorization header afterwards

### DIFF
--- a/buildkite/artifacts.go
+++ b/buildkite/artifacts.go
@@ -1,0 +1,84 @@
+// Copyright 2014 Mark Wolfe. All rights reserved.
+//
+// Use of this source code is governed by a BSD-style
+// license that can be found in the LICENSE file.
+
+package buildkite
+
+import (
+	"fmt"
+	"io"
+)
+
+// ArtifactsService handles communication with the artifact related
+// methods of the buildkite API.
+//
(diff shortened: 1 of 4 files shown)


wolfeidau:
I had a look over this and i think it is a good compromise given I didn't consider this requirement when i first wrote this library.

Need to review with @keithpitt and @sj26 how we could better address this in a way which is consistent across API wrappers.

Cheers

PR #16, opened 2017-11-23
Add retry with exp backoff for GET/429 responses

lox:
The Buildkite API will soon give HTTP 429 responses for when rate limits are hit. This retries idempotent GET requests. 

Any feels @wolfeidau? 

It brings in https://github.com/cenkalti/backoff to provide the exponential backoff with jitter.


wolfeidau:
Looks great!

PR #21, opened 2018-05-14
Implement CRUD for pipelines

haines:
Adds support for all the endpoints under `/pipelines`.

I've added structs for the provider settings but I'm not sure what the API offers for things like GitHub Enterprise. I'd appreciate some help from someone who knows!

I toyed with the idea of just returning `map[string]interface{}` for settings of unknown providers, but it was awkward to work with, so I think it's better just to return `nil`. The client doesn't currently return provider settings so I think this is a step forward in any case.


lox:
Shall we consolidate on #21?

buildkite/pipelines.go
```diff
@@ -90,3 +98,73 @@ func (ps *PipelinesService) List(org string, opt *PipelineListOptions) ([]Pipeli
...
+		return nil, resp, err
+	}
+
+	return pipeline, resp, err
+}
+
+// Edit a pipeline by slug for a given organization.
+//
+// buildkite API docs: https://buildkite.com/docs/rest-api/pipelines#update-a-pipeline
+func (ps *PipelinesService) Edit(org string, slug string, p *Pipeline) (*Pipeline, *Response, error) {
```
lox: I think this should be `Update`, unless there is prior art for `Edit` in the other endpoints.

lox:
Actually, I just dug a bit more into this one. I think I actually like the approach of typed structures for the different providers. Thoughts @mubeta06?

lox:
🤔That said, I think #22 is probably more in-line with the rest of the API and is the better incremental change. I'm going to merge that one and then we can look at ways to bring in some type safety to the `ProviderSettings` map.

haines:
Sweet, I'll raise a new PR with the provider settings.

buildkite/pipelines.go
haines: I don't think there were any `Update` or `Edit` endpoints anywhere in this client yet, and personally I prefer `Update` too, but I was going off what `go-github` does 🤷‍♂️

PR #20, opened 2018-04-15
Create pipeline

mubeta06:
And some other verb support i.e. Get and Delete.

PR #22, opened 2018-05-15
Pipeline CRUD

mubeta06:
@lox this may well supersede my previous PR (https://github.com/buildkite/go-buildkite/pull/20)

As I mentioned this work is to support the Buildkite Terraform provider we are developing (https://github.com/COzero/terraform-provider-buildkite)

### DIFF
--- a/buildkite/pipelines.go
+++ b/buildkite/pipelines.go
@@ -15,6 +15,24 @@ type PipelinesService struct {
 	client *Client
 }
 
+// Create a Pipeline.
+type CreatePipeline struct {
+	Name       string `json:"name"`
+	Repository string `json:"repository"`
+	Steps      []Step `json:"steps"`
+
+	// Optional fields
+	Description                     string            `json:"description,omitempty"`
+	Env                             map[string]string `json:"env,omitempty"`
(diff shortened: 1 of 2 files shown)


buildkite/pipelines.go
```diff
@@ -15,6 +18,24 @@ type PipelinesService struct {
...
+	Steps      []Step `json:"steps"`
+
+	// Optional fields
+	Description                     string            `json:"description,omitempty"`
+	Env                             map[string]string `json:"env,omitempty"`
+	ProviderSettings                map[string]bool   `json:"provider_settings,omitempty"`
+	BranchConfiguration             string            `json:"branch_configuration,omitempty"`
+	SkipQueuedBranchBuilds          bool              `json:"skip_queued_branch_builds,omitempty"`
+	SkipQueuedBranchBuildsFilter    string            `json:"skip_queued_branch_builds_filter,omitempty"`
+	CancelRunningBranchBuilds       bool              `json:"cancel_running_branch_builds,omitempty"`
```
haines: doesn't this mean that `false` is always omitted from the JSON?

PR #23, opened 2018-05-17
Add typed provider settings

haines:
Extracted from #21, this adds types for the different providers' settings. As before, I'd appreciate some help from someone who knows what the API offers for things like GitHub Enterprise.

### DIFF
--- a/buildkite/pipelines.go
+++ b/buildkite/pipelines.go
@@ -27,7 +27,7 @@ type CreatePipeline struct {
 	// Optional fields
 	Description                     string            `json:"description,omitempty"`
 	Env                             map[string]string `json:"env,omitempty"`
-	ProviderSettings                map[string]bool   `json:"provider_settings,omitempty"`
+	ProviderSettings                ProviderSettings  `json:"provider_settings,omitempty"`
 	BranchConfiguration             string            `json:"branch_configuration,omitempty"`
 	SkipQueuedBranchBuilds          bool              `json:"skip_queued_branch_builds,omitempty"`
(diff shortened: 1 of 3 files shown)


weirdtales:
Please look at merging this.

lox:
Apologies, I forgot to follow up on this one. 

I'm a bit hesitant to proceed with the interface-based approach, I don't think polymorphism does us any favours here as you'd always need to cast it to specific subtypes. 

I reckon we'd be better off having specific props in the `Pipeline` struct, e.g:

```go
// Pipeline represents a buildkite pipeline.
type Pipeline struct {
	...

	// the provider settings for the various supported providers
	GitlabProvider *GitlabProvider `json:"-"`
	GithubProvider *GithubProvider `json:"-"`
        GithubEnterpriseProvider *GithubProvider `json:"-"`
        ...

	// build steps
	Steps []*Step `json:"steps,omitempty"`
}
```

lox:
Oh, just looking at the code again, I think it would be more a case of typed `Settings` objects in the `Provider`.

nhooyr:
The advantage to polymorphism is that it more accurately reflects the structure of the object in that there can only be one provider settings.

lox:
> The advantage to polymorphism is that it more accurately reflects the structure of the object in that there can only be one provider settings.

Yup, I agree, but I still think it's less go-like in that you end up having to cast as part of regular usage. 

I'm not that fussed, I'd take this as-is vs not having it at all.

lox:
I decide to merge it. Sorry for the long, long delay @haines!

PR #25, opened 2018-06-29
Parse AgentQueryRules and TimeoutInMinutes for Steps

lox:
This provides some types for `AgentQueryRules` and `TimeoutInMinutes`.

@keithpitt / @sj26 can you think of any times that these would not be nilable `[]string` and `int` respectively? 

Extracted from https://github.com/buildkite/go-buildkite/pull/23


keithpitt:
@lox seems fine to me

PR #27, opened 2018-09-04
Added support for rebuilding builds

kevsmith:
We're using go-buildkite as part of a CLI tool and needed support for triggering rebuilds. This PR adds a `Rebuild()` function to `BuildsService`.


lox:
Sorry for the delay @kevsmith! Thanks for this!

PR #36, opened 2019-02-19
Update comments to have func name, run gofmt and golint

charlottestjohn:
Just recently starting using this client; and figured I might clean some things up! :)

What this PR does: 

1. Runs gofmt on the files
2. Fixes complaints from golint
except for this one: 

```
buildkite.go:99:6: func SetHttpDebug should be SetHTTPDebug
```

because I didn't want to change the name of a function!


lox:
Thanks, much appreciated! This is a very old community contributed project, so it's a bit behind in modern golang practices. 😅

PR #37, opened 2019-02-19
Switch from govendor to go dep

charlottestjohn:
While browsing through the code base, some of the dependencies were not present in the vendor folder, so I figured I'd fix it and in attempting to fix it - figured I should make a PR to switch over to dep instead of govendor. Dep is the chosen vendor management tool, https://github.com/golang/dep and provides a clean way to switch from previous vendor management solutions. 

This PR was generated using

```
dep init
```

The old vendor folder is preserved, under the date tag. 

To update deps again, you just need to do : 

```
dep ensure
```


lox:
Thanks @charlottestjohn! We've been moving to Go Modules for lots of our repos, I think I'd prefer to consolidate that approach vs dep which is unfortunately likely to be left behind in future golang releases. 

Are y'all using modules internally at all?

charlottestjohn:
Makes sense! Closing, we're not using go modules internally here

PR #41, opened 2019-03-22
Allow updating pipeline provider settings

gdhagger:
Current implementation does not allow the client to update the provider settings of a pipeline.  This simple fix allows those modifications.


lox:
Seems extremely reasonable!

PR #44, opened 2019-03-25
Add YAML support to types

gdhagger:
This PR adds YAML unmarshalling specs to the go-buildkite types.

I discovered that when decoding a YAML file that matched the JSON specs I was missing some fields.  Adding these rules fixed that issue.


sj26:
Great, thank you!

PR #47, opened 2019-04-12
 Add filtering option for listing builds by commit

srmocher:
Adds an option to filter builds by commit which is supported by the API.
https://buildkite.com/docs/apis/rest-api/builds#list-builds-for-an-organization


lox:
Thanks, this is great!

PR #49, opened 2019-07-26
Include `source` field for the builds API response

angulito:
In the builds API response, you are sending `source` field but this library is not parsing it

This PR adds support to include the `source` field


lox:
Thanks!

PR #52, opened 2019-12-13
Add support for the Job ArtifactsURL field

aashishkapur:
Listed here: https://buildkite.com/docs/apis/rest-api/jobs#retry-a-job, but the field is not in the `Job` struct


aashishkapur:
cc @lox

keithpitt:
Thanks @aashishkapur!

PR #53, opened 2020-01-02
Add client func for the Artifacts ListByJob endpoint

aashishkapur:
This endpoint is defined in the REST API but not in the golang client. Tested this manually.


aashishkapur:
@keithpitt

PR #54, opened 2020-01-03
Add functionality for teams endpoint

pyasi:
Adding support for the Teams REST endpoint based on the functionality described [here](https://buildkite.com/docs/apis/rest-api/teams#list-teams).

### DIFF
--- a/buildkite/teams.go
+++ b/buildkite/teams.go
@@ -1,0 +1,50 @@
+// Copyright 2014 Mark Wolfe. All rights reserved.
+//
+// Use of this source code is governed by a BSD-style
+// license that can be found in the LICENSE file.
+
+package buildkite
+
+import "fmt"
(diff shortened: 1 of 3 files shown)


pyasi:
@lox Would love a review when you get a chance. No rush!

buildkite/teams.go
```diff
@@ -0,0 +1,50 @@
+// Copyright 2014 Mark Wolfe. All rights reserved.
```
pda: This copyright notice doesn't seem accurate.
Perhaps just remove it completely?

pyasi: Yeah I was following what I saw in other files and recent PRs, but agree we should remove it. Thanks!

pda:
Thanks @pyasi :pray:

PR #55, opened 2020-01-13
License/Copyright tidy

pda:
Remove per-source-file copyright message; it's redundant with the `LICENSE` file. Most source files have contributions from many authors. At least one file (`artifacts.go`) had no contributions from the listed copyright holder; presumably copy/paste/modify/commit. This is likely to keep happening and get confusing (e.g. #54). I don't think we gain anything by stating/tracking copyright separately in each file. It's in the project-level LICENSE file.

Also, updating the Copyright range in LICENSE from 2019 to 2020.

(Ping @wolfeidau - let me know if you had any strong reasons/opinions for per-file notices)

### DIFF
--- a/LICENSE
+++ b/LICENSE
@@ -1,4 +1,4 @@
-Copyright (c) 2015-2019, Mark Wolfe, Buildkite Pty Ltd
+Copyright (c) 2015-2020, Mark Wolfe, Buildkite Pty Ltd
 All rights reserved.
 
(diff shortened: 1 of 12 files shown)


wolfeidau:
@pda I am happy with the change, to be honest I was following a pattern used in some of the google operated projects at the time. I have gone back to just a LICENSE file in the repository. 

Thanks for the ping. 👍

PR #56, opened 2020-01-13
Go modules; delete vendor/

pda:
Use [Go Modules](https://github.com/golang/go/wiki/Modules) instead of committing `vendor/`

The `go.mod` file was derived from previous `vendor/` content by `go mod init`.

PR #63, opened 2020-02-19
Go Modules: /v2 suffix on module path

pda:
This was missed when [initially migrating this project to Go Modules](https://github.com/buildkite/go-buildkite/pull/56); the library was already at v2.x so should have had a version suffix added to the module path. [The docs](https://github.com/golang/go/wiki/Modules#releasing-modules-v2-or-higher) suggest incrementing the major version while migrating a v2+ project to Go Modules, which would have landed us on v3.0.0. However we have now already [published a v2.x tag](https://github.com/buildkite/go-buildkite/releases/tag/v2.3.0) with Go Modules, so we're going to stick with v2 major version for now.

I've also snuck a CHANGELOG.md into this PR, generated by https://github.com/buildkite/ghch

After this merges:

- [ ] Prepare and tag a v2.3.1 release, including CHANGELOG.md update

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,0 +1,64 @@
+## [v2.3.0](https://github.com/buildkite/go-buildkite/tree/v2.3.0) (2020-02-18)
+[Full Changelog](https://github.com/buildkite/go-buildkite/compare/v2.2.0...v2.3.0)
+
+### Changed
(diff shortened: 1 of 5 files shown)


pda:
> I'm still learning the ins and outs of go modules

Turns out we all are :)

PR #65, opened 2020-03-27
add stop agent method

pmenglund:


pmenglund:
@lox ☝️

lox:
Thanks for this! Sorry for the delay!

PR #67, opened 2020-05-08
pipelines: add env struct

stephen:
The rest API sends env back, but the go struct does not expose it.

https://buildkite.com/docs/apis/rest-api/pipelines


lox:
Thanks!

Issue #66, opened 2020-03-30
Unable to upgrade to v2.3.1


Matt-Hurd:
When attempting to follow the usage, I'm getting the following issue:
```
> go get github.com/buildkite/go-buildkite/buildkite/v2
go get github.com/buildkite/go-buildkite/buildkite/v2: module github.com/buildkite/go-buildkite@upgrade found (v2.2.0+incompatible), but does not contain package github.com/buildkite/go-buildkite/buildkite/v2
```

When attempting to vendor with a manually updated `go.mod`, it looks like `@latest` hasn't been tagged with `v2.3.1`
```
> go mod vendor
go: finding module for package github.com/buildkite/go-buildkite/buildkite/v2
github.com/.../.../pkg/buildkite imports
	github.com/buildkite/go-buildkite/buildkite/v2: module github.com/buildkite/go-buildkite@latest found (v2.2.0+incompatible), but does not contain package github.com/buildkite/go-buildkite/buildkite/v2
```

elliebike:
Running `go get github.com/buildkite/go-buildkite/v2` works OK instead

I'm not sure if the `go.mod` or the `README` is wrong here, but one needs correcting!

lox:
Sorry about that folks!

pda:
Seems to be fixed in #68 and #69 :)

PR #68, opened 2020-05-12
Update README.md with correct go get

elliebike:
The command listed here doesn't actually work, see issue #66


lox:
Oooops! Thanks!

PR #69, opened 2020-06-05
Update README.md with correct import path

justcompile:
The import path in the Readme was incorrect as per https://github.com/buildkite/go-buildkite/pull/68 :)


pda:
Thanks 😊

PR #70, opened 2020-08-18
buildkite: support GitHub Enterprise settings

stevvooe:
This change adds "github_enterprise" to the switch when resolving the 
settings type during unmarshaling. I'm not sure if we have the same 
settings between GitHub and GitHub Enterprise but this should unblock 
simple use cases for now.

There might be a better fix for this, so let me know if I'm missing anything.

### DIFF
--- a/buildkite/providers.go
+++ b/buildkite/providers.go
@@ -27,7 +27,10 @@ func (p *Provider) UnmarshalJSON(data []byte) error {
 	switch v.ID {
 	case "bitbucket":
 		settings = &BitbucketSettings{}
-	case "github":
+	case "github", "github_enterprise":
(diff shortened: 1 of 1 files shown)


pda:
Hey @stevvooe, thanks! I think this would pretty much work, but the list of supported settings are [slightly different](https://buildkite.com/docs/apis/rest-api/pipelines#provider-settings-properties); so @niceking is working on a patch that introduces a GitHubEnterpriseSettings struct with a subset of the GitHubSettings; it will include `build_tags` which I believe is the one you're after.

stevvooe:
@pda Thanks! I figured they would be slightly different. I'll close this.

PR #71, opened 2020-08-19
Add github enterprise settings

niceking:
GitHub Enterprise settings are missing from our provider list, have updated the settings to match the properties specified in our [REST API](https://buildkite.com/docs/apis/rest-api/pipelines#provider-settings-properties) documentation.

### DIFF
--- a/buildkite/providers.go
+++ b/buildkite/providers.go
@@ -29,6 +29,8 @@ func (p *Provider) UnmarshalJSON(data []byte) error {
 		settings = &BitbucketSettings{}
 	case "github":
 		settings = &GitHubSettings{}
+	case "github_enterprise":
+		settings = &GitHubEnterpriseSettings{}
 	case "gitlab":
 		settings = &GitLabSettings{}
 	default:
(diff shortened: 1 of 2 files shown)


pda:
Nice, @niceking 👑

Although maybe we should keep the `bump go version` commit separate from this PR?

niceking:
Yeahh cool I've removed that commit and discovered a newer version of Go and made another PR #72

pda:
Hey @stevvooe, does this look like it achieves what you were after in #70?

buildkite/providers.go
```diff
@@ -84,6 +86,23 @@ type GitHubSettings struct {
 
 func (s *GitHubSettings) isProviderSettings() {}
 
+
```
pda, line 89: Minor detail; `gofmt` doesn't seem to want this extra line. You might want to add https://atom.io/packages/go-plus or similar to your Atom setup, so that `gofmt` or `goimports` runs automatically.

pda (approved):
Looking good to me 👌🏼

stevvooe:
@pda Yes, this looks great! Thank you!

PR #73, opened 2020-09-08
Format slug correctly for Pipeline GET

ellie:
The Buildkite API will replace "/" with "-" to create a slug, when the
name of a pipeline contains a "/". We should do that here as well to
prevent usage errors :)


JuanitoFatas:
The client will get a failure if it requested the wrong thing and we don’t want to replicate the slug generation rules here. I’m going to close this. Much appreciate you sent this through and please send more in the future. Cheers 💚

PR #74, opened 2020-09-24
Add missing fields to structs

kushmansingh:
There fields specified in the API docs that are missing from this client lib, added them after verifying they existed via curl

### DIFF
--- a/buildkite/pipelines.go
+++ b/buildkite/pipelines.go
@@ -180,22 +180,7 @@ func (ps *PipelinesService) Update(org string, p *Pipeline) (*Response, error) {
 
 	u := fmt.Sprintf("v2/organizations/%s/pipelines/%s", org, *p.Slug)
 
-	// There is quite a lot of properties that are not represented by the Client-side
-	// Pipeline abstraction hence only a subset can be updated.
-	cp := &CreatePipeline{
-		Name:       *p.Name,
-		Repository: *p.Repository,
-		Steps:      make([]Step, len(p.Steps)),
-	}
-	if p.Provider != nil {
-		cp.ProviderSettings = p.Provider.Settings
-	}
-
-	for i := range p.Steps {
-		cp.Steps[i] = *p.Steps[i]
-	}
-
-	req, err := ps.client.NewRequest("PATCH", u, cp)
+	req, err := ps.client.NewRequest("PATCH", u, p)
 	if err != nil {
 		return nil, err
 	}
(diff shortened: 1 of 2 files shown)


buildkite/pipelines.go
```diff
@@ -20,6 +20,8 @@ type CreatePipeline struct {
 	Steps      []Step `json:"steps" yaml:"steps"`
 
 	// Optional fields
+	Configuration                   string            `json:"configuration,omitempty" yaml:"configuration,omitempty"`
```
ticky, line 23: With YAML steps enabled, you supply either Steps or Configuration, so I believe these both need to be optional now:

```suggestion

	// Optional fields
	Steps                           []Step            `json:"steps,omitempty" yaml:"steps,omitempty"`
	Configuration                   string            `json:"configuration,omitempty" yaml:"configuration,omitempty"`
```

buildkite/pipelines.go
```diff
@@ -168,22 +178,7 @@ func (ps *PipelinesService) Update(org string, p *Pipeline) (*Response, error) {
...
-	if p.Provider != nil {
-		cp.ProviderSettings = p.Provider.Settings
-	}
-
-	for i := range p.Steps {
-		cp.Steps[i] = *p.Steps[i]
-	}
-
-	req, err := ps.client.NewRequest("PATCH", u, cp)
+	req, err := ps.client.NewRequest("PATCH", u, p)
```
ticky, line 181: I think this works, I’m not sure what properties that comment was talking about. 👍🏼

ticky (changes requested):
Looking great, I think there’s one change which would be helpful

sj26:
Neat!

Will this break if we add new fields to these responses?

buildkite/pipelines.go
```diff
@@ -17,9 +17,13 @@ type PipelinesService struct {
 type CreatePipeline struct {
 	Name       string `json:"name" yaml:"name"`
 	Repository string `json:"repository" yaml:"repository"`
-	Steps      []Step `json:"steps" yaml:"steps"`
+
+	// Either configuration needs to be specified as a yaml string or steps.
+	Configuration string `json:"configuration,omitempty" yaml:"configuration,omitempty"`
+	Steps         []Step `json:"steps" yaml:"steps"`
```
sj26, line 23: Should this be modified to `omitempty` as well?

jayco, line 23:
```suggestion
	Steps         []Step `json:"steps, omitempty" yaml:"steps"`
```

JuanitoFatas:
@kushmansingh Thanks Kush so much for your contributions ❤️  I will cut a release soon.

PR #77, opened 2020-12-14
Add soft_failed attribute for job

qinjin:
This is part of the Jobs API model, and having it would be useful for people to know whether a job in the [build](https://buildkite.com/docs/apis/rest-api/builds#get-a-build) is `soft_failed` or not.


sj26:
Thanks!

PR #82, opened 2021-05-11
Add support for plugins to Pipeline steps

luma:
I needed to create pipelines that use the docker-compose plugin but it wasn't supported yet.

I've updated the Pipeline tests to include the plugins where ever I could.


jayco:
Sorry for the delay in reviewing this, thanks for contribution @luma 🎉

PR #84, opened 2021-06-08
Adding cluster_id fields to Pipeline and CreatePipeline structs

ksepehr:
These fields were missing from go-buildkite. Without these fields it's impossible to use the cluster feature along with this convenient client library. In our use-case, this was for creating pipelines in a given cluster and identifying which cluster pipelines belong to.

I have tested deploying and viewing clusters in my organization and the new fields do indeed work.


ksepehr:
Looks like the cause of failure is unrelated to the PR itself.

image

jayco (approved):
Looks good, thanks @ksepehr 🎉

PR #87, opened 2021-07-19
Implement Access Tokens API

alam0rt:
Simply implemented the API as per https://buildkite.com/docs/apis/rest-api/access-token


buildkite/access_tokens.go
```diff
@@ -0,0 +1,62 @@
...
+		return nil, resp, err
+	}
+
+	return accessToken, resp, nil
+}
+
+// Revokes the current token which was used to authenticate the request
+//
+// buildkite API docs: https://buildkite.com/docs/rest-api/access-token
+func (ats *AccessTokensService) Revoke() (*AccessToken, *Response, error) {
```
jayco, line 42: When I ran this, it returned an error:

```
➜  go-buildkite git:(alam0rt/master) ✗ go run test.go token=$JJ_BK_TOKEN 
EOF2021/07/26 16:32:54 json encode failed: EOF
exit status 1
```

I think the issue is that the API is returning a 204 there [revoke-the-current-token](https://buildkite.com/docs/apis/rest-api/access-token#revoke-the-current-token) and the JSON decoder is complaining [here](https://github.com/buildkite/go-buildkite/blob/master/buildkite/buildkite.go#L273)?

Since it's a Delete call, maybe it would be better to do something similar to [delete-an-agent](https://github.com/buildkite/go-buildkite/blob/master/buildkite/agents.go#L127)?

I quickly played around and came up with something like:
```suggestion
func (ats *AccessTokensService) Revoke() (*Response, error) {

    var u string

    u = fmt.Sprintf("v2/access-token")

    req, err := ats.client.NewRequest("DELETE", u, nil)
    if err != nil {
        return nil, err
    }

    resp, err := ats.client.Do(req, nil)
    if err != nil {
        return resp, err
    }

    return resp, nil
}
```

buildkite/access_tokens_test.go
```diff
@@ -0,0 +1,56 @@
...
+	}
+}
+
+func TestAccessTokensService_Revoke(t *testing.T) {
+	setup()
+	defer teardown()
+
+	mux.HandleFunc("/v2/access-token", func(w http.ResponseWriter, r *http.Request) {
+		testMethod(t, r, "DELETE")
+		fmt.Fprint(w, `{}`)
```
jayco, line 41: As mentioned in the comment, I think this status is a StatusNoContent.

I haven't tested this but I think you could return something like:
```suggestion
		w.WriteHeader(http.StatusNoContent)
```

buildkite/access_tokens_test.go
```diff
@@ -0,0 +1,56 @@
...
+		testMethod(t, r, "GET")
+		fmt.Fprint(w, `{"uuid": "b63254c0-3271-4a98-8270-7cfbd6c2f14e","scopes": ["read_build"]}`)
+	})
+
+	ats, _, err := client.AccessTokens.Get()
+	if err != nil {
+		t.Errorf("AccessTokens.Get returned error: %v", err)
+	}
+
+	fmt.Print(ats)
```
jayco, line 24: We could probably remove this one?

```suggestion
	
```

buildkite/access_tokens_test.go
```diff
@@ -0,0 +1,56 @@
...
+		testMethod(t, r, "DELETE")
+		fmt.Fprint(w, `{}`)
+	})
+
+	ats, _, err := client.AccessTokens.Revoke()
+	if err != nil {
+		t.Errorf("AccessTokens.Revoke returned error: %v", err)
+	}
+
+	fmt.Print(ats)
```
jayco, line 49: and this one?
```suggestion

```

buildkite/access_tokens.go
```diff
@@ -0,0 +1,62 @@
...
+	req, err := ats.client.NewRequest("GET", u, nil)
+	if err != nil {
+		return nil, nil, err
+	}
+
+	accessToken := new(AccessToken)
+
+	resp, err := ats.client.Do(req, accessToken)
+	if err != nil {
+		fmt.Print(err)
```
jayco, line 32: Should we remove this one?

```suggestion

```

jayco (changes requested):
Thanks @alam0rt for the awesome PR! 💪

Just a couple of suggestions here, let me know what you think?

buildkite/access_tokens.go
alam0rt: Have implemented this suggestion, thanks :)

buildkite/access_tokens_test.go
alam0rt: Have done this and tweaked the test to not touch body as it was causing EOF

buildkite/access_tokens.go
alam0rt: done

alam0rt:
> Thanks @alam0rt for the awesome PR! 
> 
> Just a couple of suggestions here, let me know what you think?

Thanks for the suggestions, all totally valid and I did miss a few things!

All implemented now :+1:

jayco (approved):
Awesome, thanks for the contribution @alam0rt 🎉

PR #89, opened 2021-08-17
Fix pagination for teams 

atishpatel:
The teams API has pagination but this library is missing pagination for teams.


jayco:
Thanks for the contribution @atishpatel 🎉

PR #90, opened 2021-08-19
Add block-job label

BEvgeniyS:
Block label is missing, which makes it harder to target in cases where pipeline doesn't define it


jayco:
Thanks for the contribution @BEvgeniyS 🎉

PR #91, opened 2021-08-20
Version bump to V3

jayco:
There was a breaking change in https://github.com/buildkite/go-buildkite/pull/89 so I have done the things needed for a major version bump.

- [x] update import paths
- [x] update version
- [x] update changelog
- [x] go version minor bump
- [x] go mod tidy

### DIFF
--- a/go.mod
+++ b/go.mod
@@ -1,6 +1,6 @@
-module github.com/buildkite/go-buildkite/v2
+module github.com/buildkite/go-buildkite/v3
 
-go 1.15
+go 1.16
 
 require (
 	github.com/alecthomas/template v0.0.0-20190718012654-fb15b899a751 // indirect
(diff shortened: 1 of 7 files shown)


go.mod
```diff
@@ -1,6 +1,6 @@
-module github.com/buildkite/go-buildkite/v2
+module github.com/buildkite/go-buildkite/v3
 
-go 1.15
+go 1.16
```
keithduncan, line 3: This bump [drops support for macOS 10.11](https://github.com/buildkite/agent/blob/main/docs/macos.md) which is likely fine I just wanted to note it 😄 we don’t support 10.11 in the agent any more either

keithduncan (approved):
Looks good to me, 100% your call on whether to do a major bump vs a backwards compatible change

go.mod
jayco: Good to know!

jradtilbrook (approved):
Ohh exciting! LGTM 👍

PR #95, opened 2021-12-10
Add clean_checkout to CreateBuild struct

justingallardo-okta:
This adds the `clean_checkout` field to the `CreateBuild` struct so that clients can create builds with a clean checkout. API documentation available at https://buildkite.com/docs/apis/rest-api/builds#create-a-build


lizrabuya (approved):
Nice!

PR #96, opened 2021-12-11
Support filtering list builds call by multiple branches

gsavit:
Per [docs](https://buildkite.com/docs/apis/rest-api/builds#list-builds-for-a-pipeline), we should be able to filter by more than one branch.


lizrabuya (approved):
works for me!

PR #102, opened 2022-10-17
update teams docs

sfunkhouser:
typo that had user when it should be team


lizrabuya (approved):
Good catch. Thank you!

PR #104, opened 2022-12-08
add group_key to Job struct if available

alexnguyennn:
### Background
The [Get Build](https://buildkite.com/docs/apis/rest-api/builds#get-a-build) rest api now returns a `group_key` property on jobs if they are part of a [group step](https://buildkite.com/docs/pipelines/group-step). However, the Get Job itself doesn't yet include this property (confirmed via support email).

go-buildkite's `Job` struct doesn't yet include this property and it would be useful to include this to allow SDK usecases that look up jobs belonging to a specific group.

### What's changed
* add `GroupKey` to `Job` struct and omit if missing like other properties
* update `Get` Build test to handle multiple scenarios and add one that includes a job with `group_key` in return payload

### How has this been tested?

- Confirmed unit tests still pass
- Tested working by consuming fork changes and confirming the `group_key` property is parsed properly from the Get Build REST API response.

Happy to make any changes required before merging, thanks in advance!

### DIFF
--- a/buildkite/builds_test.go
+++ b/buildkite/builds_test.go
@@ -53,23 +53,56 @@ func TestBuildsService_List(t *testing.T) {
 }
 
 func TestBuildsService_Get(t *testing.T) {
-	setup()
-	defer teardown()
+	buildNumber := "123"
+	orgName := "my-great-org"
+	pipelineName := "sup-keith"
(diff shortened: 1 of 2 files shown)


lizrabuya:
hey @alexnguyennn , we just released v3.2.0, and these changes are causing conflicts. Would you be able to update them with the latest version?

PR #110, opened 2023-01-20
Improve README grammar

mdb:
👋 Hello - This PR seeks to correct a few grammar issues in the `README`:

1. correctly use the possessive "its" rather than the contraction "it's"
2. Reword run-on sentences

As a bonus, it also makes some of the README texts links, which may be helpful for readers.

Thanks!


lizrabuya (approved):
LGTM!

PR #100, opened 2022-05-04
Add `GraphqlID` field to `Build`, `Job` and `Pipeline`.

fatmcgav:
The `graphql_id` JSON field can be useful for downstream usage, so add
the field to the relevant structs to expose it.

PR #116, opened 2023-01-23
Add GraphQLID to more api objects

benmoss:
I need it on pipelines specifically, but I figured I'd add it to a couple more objects just for the sake of future users


benmoss:
Ah, just saw #100 overlaps a bit with this, merge whichever you want 😄

lizrabuya (approved):
Looks good.

Issue #118, opened 2023-02-27
Job struct is missing `parallel_group_index` and `parallel_group_total`


yqiu24:
These fields are missing in the Job struct, even though they exist in the Jobs API docs: https://buildkite.com/docs/apis/rest-api/jobs

yqiu24:
Created a PR for this, could someone take a look? https://github.com/buildkite/go-buildkite/pull/119

lizrabuya:
@yqiu24 , this will be part of a patch release 3.2.1. Thanks

PR #119, opened 2023-03-01
Add missing parallel job fields to struct

yqiu24:
Add the `parallel_group_index` and `parallel_group_total` fields to the `Job` struct. These fields exist in the jobs API defined [here](https://buildkite.com/docs/apis/rest-api/jobs).

Issue: https://github.com/buildkite/go-buildkite/issues/118


jradtilbrook (approved):
This looks good. @lizrabuya can you also confirm and sort out a release for these changes?

PR #124, opened 2023-04-04
Upgrade to latest go

mcncl:
Go is currently on `1.20`. It makes sense to use this version rather than upgrading to a version that'll also be outdated soon.


lizrabuya (approved):
Nice!

PR #127, opened 2023-04-07
Allow filtering builds by meta_data

andrewhamon:
I noticed that it is not possible to filter builds by meta_data, even though [this is available in the API](https://buildkite.com/docs/apis/rest-api/builds#list-all-builds).

At Flexport, we make use of some very generic pipelines, so being able to filter by metadata is very important.

The encoding used for meta_data is not natively supported by go-querystring, but they have [the Encoder interface](https://pkg.go.dev/github.com/google/go-querystring/query#Encoder) available as an escape hatch for arbitrary encoding rules

### DIFF
--- a/buildkite/builds.go
+++ b/buildkite/builds.go
@@ -2,6 +2,7 @@ package buildkite
 
 import (
 	"fmt"
+	"net/url"
 	"time"
 )
 
@@ -27,7 +28,7 @@ type CreateBuild struct {
 
 	// Optional fields
 	Author                      Author            `json:"author,omitempty" yaml:"author,omitempty"`
-	CleanCheckout 				bool 			  `json:"clean_checkout,omitempty" yaml:"clean_checkout,omitempty"`
+	CleanCheckout               bool              `json:"clean_checkout,omitempty" yaml:"clean_checkout,omitempty"`
 	Env                         map[string]string `json:"env,omitempty" yaml:"env,omitempty"`
 	MetaData                    map[string]string `json:"meta_data,omitempty" yaml:"meta_data,omitempty"`
 	IgnorePipelineBranchFilters bool              `json:"ignore_pipeline_branch_filters,omitempty" yaml:"ignore_pipeline_branch_filters,omitempty"`
@@ -84,6 +85,20 @@ type Build struct {
 	PullRequest *PullRequest `json:"pull_request,omitempty" yaml:"pull_request,omitempty"`
 }
 
+type MetaDataFilters struct {
+	MetaData map[string]string
+}
+
+// Encodes MetaData in the format expected by the buildkite API
+// Example: ?meta_data[some-key]=some-value
+func (mo MetaDataFilters) EncodeValues(parent_key string, v *url.Values) error {
+	for key, val := range mo.MetaData {
+		keyString := fmt.Sprintf("%s[%s]", parent_key, key)
+		v.Add(keyString, val)
+	}
+	return nil
+}
+
 // BuildsListOptions specifies the optional parameters to the
 // BuildsService.List method.
 type BuildsListOptions struct {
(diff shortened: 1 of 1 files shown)


buildkite/builds.go
```diff
@@ -27,7 +28,7 @@ type CreateBuild struct {
 
 	// Optional fields
 	Author                      Author            `json:"author,omitempty" yaml:"author,omitempty"`
-	CleanCheckout 				bool 			  `json:"clean_checkout,omitempty" yaml:"clean_checkout,omitempty"`
+	CleanCheckout               bool              `json:"clean_checkout,omitempty" yaml:"clean_checkout,omitempty"`
```
andrewhamon, line 31: gofmt did this, let me know if you'd like me to change it back

james2791:
Hey @andrewhamon!

Thanks alot for your contribution! Would you be able to get some evidence of the filtering by certain meta-data in action (without any sensitive tokens etc)? Will try from my end also but best to get concrete proof to then merge this in.

Thanks 👍

buildkite/builds.go
james2791: Should be OK - looks fine with me

andrewhamon:
Hey @james2791, absolutely can do that. Sorry for the delay, I've been on a long vacation. Hopefully I can get an example working soon :)

james2791:
> Hey @james2791, absolutely can do that. Sorry for the delay, I've been on a long vacation. Hopefully I can get an example working soon :)

Thanks a million @andrewhamon 👍 

Just helps with keeping our examples repo up to date and with the newer functionality!

james2791:
Hey @andrewhamon!

Wondering if you got an example to share for this metadata filtering functionality? Did a quick look into adding the `MetaDataFilters` struct into an example but it seems like the URL constructed back doesn't yield the expected case. For example with the below `MetaDataFilters` passed into BuildListOptions (in a pipeline builds retrieval) - the URL generated yields `/builds?MetaData%5Brelease-version%5D=2.0`

```
md := MetaDataFilters{
  MetaData: map[string]string{"release-version": "2.0"},
}
```
Cheers 😄

buildkite/builds.go
```diff
@@ -84,6 +85,20 @@ type Build struct {
...
+type MetaDataFilters struct {
+	MetaData map[string]string
+}
+
+// Encodes MetaData in the format expected by the buildkite API
+// Example: ?meta_data[some-key]=some-value
+func (mo MetaDataFilters) EncodeValues(parent_key string, v *url.Values) error {
+	for key, val := range mo.MetaData {
+		keyString := fmt.Sprintf("%s[%s]", parent_key, key)
+		v.Add(keyString, val)
```
mcncl, line 97: In here v is being added to as a url.Values list, which will perform URL "friendly" encoding when parsed. The value of this will need to be converted to a string in both the Get and List functions in order to force it to not use the URL encoded values for square brackets.

buildkite/builds.go
```diff
@@ -113,6 +128,9 @@ type BuildsListOptions struct {
 	// Include all retried jobs in each build’s jobs list
 	IncludeRetriedJobs bool `url:"include_retried_jobs,omitempty"`
 
+	// Filters results by metadata.
+	MetaData MetaDataFilters
```
james2791, line 132: @andrewhamon I believe what we would require here (after testing) would be:

```
MetaData MetaDataFilters `url:"meta_data,omitempty"`
```

By default - the [encoded](https://github.com/google/go-querystring/blob/master/query/encode.go#L57-L74) URL key when calling `encodeValues` defaults to the struct's name, unless specified with the `url` tag value as above. In our case - `MetaData` is passed as the URL parameter (which of course needs to be `meta_data` to call the builds API successfully with that parameter!

mcncl (approved):

james2791 (approved):

james2791:
@andrewhamon just letting you know that I've made the change that sees this through, you still should be listed as the contributor to the PR but can confirm this works beautifully now. 

Thanks for the contribution! 🙂

Issue #79, opened 2021-03-23
Author field (and others) are missing from the builds structure


dan-embark:
We want to pick out the Author of a build since Creator is only correct if the author has a Buildkite user.

If you look at https://github.com/buildkite/go-buildkite/blob/6c237667c3e8f373935ac89b3ed0876dbff85ef2/buildkite/builds.go#L55, it references the `Creator` with a `*Creator` type, but there is no `Author` field.

The JSON event that Buildkite sends *does* contain the Author...

Here is a sample event that Buildkite sends copied (and censored) from the Builkite Webhook Notification page:

It has the `Author` field in there already and it will only match if the email in Author matches an existing user in Buildkite with the same email.

```
{
  "event": "build.finished",
  "build": {
    "id": "xxxxxxxxxxxxxxxxxxxxxxxxxx",
    "graphql_id": "xxxxxxxxxxxxxxxxxxxxxxxxxx=",
    "number": 20125,
    "state": "passed",
    "blocked": false,
    "blocked_state": "",
    "message": "xxxxxxxxxxxxxxxxxxxxxxxxxx",
    "commit": "xxxxxxxxxxxxxxxxxxxxxxxxxx",
    "branch": "xxxxxxxxxxxxxxxxxxxxxxxxxx",
    "tag": null,
    "source": "api",
    "author": {
      "name": "n.n",
      "email": "n-n@n.com"
    },
    "creator": {
      "id": "xxxxxxxxxxxxxxxxxxxxxxxxxx",
      "graphql_id": "xxxxxxxxxxxxxxxxxxxxxxxxxx",
      "name": "xxxxxxxxxxxxxxxxxxxxxxxxxx",
      "email": "xxxxxxxxxxxxxxxxxxxxxxxxxx",
      "avatar_url": "xxxxxxxxxxxxxxxxxxxxxxxxxx",
      "created_at": "2019-12-16T14:11:42.986Z"
    },
    "created_at": "2021-03-23T13:21:52.185Z",
    "scheduled_at": "2021-03-23T13:21:52.109Z",
    "started_at": "2021-03-23T13:22:00.000Z",
    "finished_at": "2021-03-23T13:56:55.000Z",
    "meta_data": {
      "ci": "True",
      "change": "xxxxxxxxxxxxxxxxxxxxxxxxxx",
      "message": "xxxxxxxxxxxxxxxxxxxxxxxxxx",
      "fullbuild": "True",
      "is_tryjob": "true",
    },
[...]

PR #129, opened 2023-04-14
Add `rebuilt_from` field to Build struct

matthiasr:
These are present in the API, but currently undocumented, see buildkite/docs#2032. As per the community Slack, the field is stable and can be relied upon.

See also: #79


james2791:
Hello @matthiasr!

Thanks for your contribution: whipped up a small test file with the relevant API token/org and can confirm this is working (also checked out Community Slack also).

image

james2791 (approved):
LGTM 👍

PR #131, opened 2023-04-18
Added group_key to Job struct, Builds test updates

james2791:
_**NB - Raised with contributing commit vs my own one**_

Added the `GroupKey` parameter to the Job struct (`group_key` in output) - which states the Job's group if they are such defined in one as part of a [group step](https://buildkite.com/docs/pipelines/group-step).

Tested locally associating with a Build that has a group defined, notably with the following steps (sequential group command echoes):

```
steps:
  - group: ":lock_with_ink_pen: Say hello with groups"
    key: "echoes"
    steps:
      - label: ":wave: 1"
        key: wave_1
        command: "echo 'Hello 1'"
        agents:
          local: true
      - label: ":wave: 2"
        key: wave_2
        command: "echo 'Hello 2'"
        depends_on:
        - wave_1
        agents:
          local: true
      - label: ":wave: 3"
        key: wave_3
        command: "echo 'Hello 3'"
        depends_on:
        - wave_2
        agents:
          local: true
      - label: ":wave: 4"
        key: wave_4
        command: "echo 'Hello 4'"
        depends_on:
        - wave_3
        agents:
          local: true
```
The corresponding run of the `ListByPipeline` [function](https://github.com/buildkite/go-buildkite/blob/main/buildkite/builds.go#L248) gives all such builds of this pipeline: and notably each `job` inside the results shows such key as expected:

image

Also, the rewritten `TestBuildsService_Get` function with the two seperate functions pass as expected - one with a expected ID (123) and the other with a `group_key` of `job_group`

image

PR #132, opened 2023-04-19
add group_key to job struct if available

alexnguyennn:
This is a follow up of my previous PR #104 which should fix the merge conflicts with `main` that are still in #131. Thanks @lizrabuya / @james2791 for attempting to get this merged, sorry the rebase is late. :)
(Happy to close this in favour of #131 if we fix the conflict there instead, just wasn't sure if you were still waiting on me to do it)

### Background
The [Get Build](https://buildkite.com/docs/apis/rest-api/builds#get-a-build) rest api now returns a `group_key` property on jobs if they are part of a [group step](https://buildkite.com/docs/pipelines/group-step). However, the Get Job itself doesn't yet include this property (confirmed via support email).

go-buildkite's `Job` struct doesn't yet include this property and it would be useful to include this to allow SDK usecases that look up jobs belonging to a specific group.

### What's changed
* add `GroupKey` to `Job` struct and omit if missing like other properties
* update `Get` Build test to handle multiple scenarios and add one that includes a job with `group_key` in return payload

### How has this been tested?

- Confirmed unit tests still pass
- Tested working by consuming fork changes and confirming the `group_key` property is parsed properly from the Get Build REST API response.
- @james2791 's testing in #131 

Happy to make any changes required before merging, thanks in advance!

### DIFF
--- a/buildkite/builds_test.go
+++ b/buildkite/builds_test.go
@@ -53,23 +53,56 @@ func TestBuildsService_List(t *testing.T) {
 }
 
 func TestBuildsService_Get(t *testing.T) {
-	setup()
-	defer teardown()
+	buildNumber := "123"
+	orgName := "my-great-org"
+	pipelineName := "sup-keith"
+	requestSlug := fmt.Sprintf("/v2/organizations/%s/pipelines/%s/builds/%s",
+		orgName, pipelineName, buildNumber)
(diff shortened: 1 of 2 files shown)


lizrabuya (approved):
LGTM! Thanks for the contribution

james2791:
Thanks also @alexnguyennn and @lizrabuya 🙂

PR #133, opened 2023-05-04
add visibility field to pipeline create

sfunkhouser:
This will allow the `visibility` to be set on pipeline creation, currently visibility is only returned on the `get` pipeline response


lizrabuya (approved):
Nice!

PR #137, opened 2023-06-28
[SUP-1074] Request body cannot be nil

### LINKED ISSUE #136: `UnblockJob` gets a 500 server error if you pass `nil` `JobUnblockOptions`
If you pass `nil` as the last parameter to the `UnblockJob` function, the request receives a 500 Internal Server Error from Buildkite.

If you pass `&buildkite.JobUnblockOptions{ Fields: map[string]string{} }`, then go-buildkite generates the request query string `Fields=map%5B%5D`, which is invalid, but ignored by the API.

The expected behavior is that both `nil` and `&buildkite.JobUnblockOptions{ Fields: map[string]string{} }` work and generate a valid request URL.

lizrabuya:
If no unblock parameters are passed, a 500 error is returned from go-buildkite. The reason is that an null request body was passed. The expected behavior is a 400 bad request returned when no unblock parameters are passed.

This will resolve #136

### DIFF
--- a/buildkite/jobs.go
+++ b/buildkite/jobs.go
@@ -82,11 +82,6 @@ func (js *JobsService) UnblockJob(org string, pipeline string, buildNumber strin
 
 	u = fmt.Sprintf("v2/organizations/%s/pipelines/%s/builds/%s/jobs/%s/unblock", org, pipeline, buildNumber, jobID)
 
-	u, err := addOptions(u, opt)
-	if err != nil {
-		return nil, nil, err
-	}
-
 	if opt == nil {
 		opt = &JobUnblockOptions{}
 	}
--- a/buildkite/jobs.go
+++ b/buildkite/jobs.go
@@ -82,6 +82,11 @@ func (js *JobsService) UnblockJob(org string, pipeline string, buildNumber strin
 
 	u = fmt.Sprintf("v2/organizations/%s/pipelines/%s/builds/%s/jobs/%s/unblock", org, pipeline, buildNumber, jobID)
 
+	u, err := addOptions(u, opt)
+	if err != nil {
+		return nil, nil, err
+	}
+
 	if opt == nil {
 		opt = &JobUnblockOptions{}
 	}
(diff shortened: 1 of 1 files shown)


buildkite/jobs.go
```diff
@@ -87,6 +87,10 @@ func (js *JobsService) UnblockJob(org string, pipeline string, buildNumber strin
 		return nil, nil, err
```
jradtilbrook, line 87: I'm looking at https://buildkite.com/docs/apis/rest-api/jobs#unblock-a-job and that endpoint doesn't accept any query parameters. So I think lines 85-87 are redundant. It was added 5 years ago and maybe by default. The bug report in #136 mentions that query parameters are added but get ignored so I think we should actually just remove these lines as well to clean it up properly.

The rest looks good. Technically, I think this might be a bug on the server side since the `fields` body property is documented as optional. But this works too 👍

jradtilbrook (changes requested):

sj26:
> 400 bad request returned when no unblock parameters are passed

Not quite, but close — an empty json hash (`{}`) is fine, but you must provide a body to a PUT request. 👍

Omitting a body entirely currently returns a `400 Bad Request` correctly, but sending a literal JSON `null` as the request body currently causes a `500 Internal Server Error`. That will be fixed soon.

In the meantime, just make sure this api method is given an empty hash (`{}`; not `null`) as the response body when no fields or unblocker are present and you should be golden 👌

buildkite/jobs.go
lizrabuya: @jradtilbrook , I have removed the code block that appends the query parameters to the endpoint [here](https://github.com/buildkite/go-buildkite/commit/fe3c18117b6cd9cea668a34efd9b6e858332c21d ) as suggested.

lizrabuya (commented):
Looks good!

mcncl (approved):

jradtilbrook (approved):

PR #145, opened 2023-08-10
SUP-1321: Pipeline updates utilising dedicated PipelineUpdate struct

james2791:
- [x] tests added
- [x] examples of each service action (List, Get etc) added to the relevant `examples/` folder (create if new)

This PR reverts [#143](https://github.com/buildkite/go-buildkite/pull/143) while introducing the main change of allowing only-permitted properties as part of updating a pipeline via the `UpdatePipeline` struct. From the `Update` function, the passed in `Pipeline` (p) instance is used to create said `UpdatePipeline` instance with via p's values in `generateUpdatePipelineInstance`

The `Update` function's signature is also reverted to the existing implementation from `3.3.1`.

### DIFF
--- a/buildkite/pipelines.go
+++ b/buildkite/pipelines.go
@@ -39,24 +39,21 @@ type CreatePipeline struct {
 }
 
 type UpdatePipeline struct {
-	// Slug required for determining the pipeline to update
-	Slug string `json:"slug" yaml:"slug"`
-
 	// Either configuration needs to be specified as a yaml string or steps (based on what the pipeline uses)
-	Configuration string `json:"configuration,omitempty" yaml:"configuration,omitempty"`
-	Steps         []Step `json:"steps,omitempty" yaml:"steps,omitempty"`
+	Configuration string  `json:"configuration,omitempty" yaml:"configuration,omitempty"`
+	Steps         []*Step `json:"steps,omitempty" yaml:"steps,omitempty"`
 
-	Name                            string           `json:"name,omitempty" yaml:"name,omitempty"`
-	Repository                      string           `json:"repository,omitempty" yaml:"repository,omitempty"`
-	DefaultBranch                   string           `json:"default_branch,omitempty" yaml:"default_branch,omitempty"`
-	Description                     string           `json:"description,omitempty" yaml:"description,omitempty"`
+	Name                            *string          `json:"name,omitempty" yaml:"name,omitempty"`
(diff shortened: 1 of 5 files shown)


buildkite/pipelines.go
```diff
@@ -232,25 +236,24 @@ func (ps *PipelinesService) Delete(org string, slug string) (*Response, error) {
 // Update - Updates a pipeline.
 //
 // buildkite API docs: https://buildkite.com/docs/rest-api/pipelines#update-a-pipeline
-func (ps *PipelinesService) Update(org, slug string, p *UpdatePipeline) (*Pipeline, *Response, error) {
+func (ps *PipelinesService) Update(org string, p *UpdatePipeline) (*Response, error) {
```
jradtilbrook, line 239: Oh wait but this is still a breaking change isn't it? (compared to the last release at least).

I thought you mean something more like this:

```go
func (ps *PipelinesService) Update(org string, p *Pipeline) (*Response, error) {
   update := UpdatePipeline{
      Name: p.Name,
      ...
   }
...
```

So that the function footprint is compatible with the current release, but internal we start moving it towards an UpdatePipeline function argument. And we can look to make a breaking change in v4 and consolidate Update and Create so they work similarly.

UpdatePipeline struct can define all the omitempty properties so that it only marshals to json values that have been set to non-default values. But this may cause a problem for any pipeline settings that the user wants to specifically set to false, which is where the map[string]interface approach would work for I think

lizrabuya: I agree with @jradtilbrook  on this one. This is still a breaking change.  We can start applying such when we plan to do a v4 release.

james2791: Re-implemented with the above in mind 👍

jradtilbrook (approved):
I think this looks good 👍

I just want to confirm how `omitempty` works. Because the UpdatePipeline struct uses pointers, the empty value of a pointer is nil. So if the struct field is nil, it won't be sent in the JSON PATCH request. But if its set to empty strings or false then it will be sent. Is that correct?

Does that cause any issues where things want to be set to `null`?

jradtilbrook (approved):

james2791:
Correct @jradtilbrook - depending on what the property is set as, an empty string/bool is constructed based off when the request [body](https://github.com/buildkite/go-buildkite/blob/main/buildkite/buildkite.go#L127) is made. `nil`s excluded with the `omitempty` 👍

jradtilbrook (approved):
And its no longer a breaking change for the function signature which is great! 🙌

PR #156, opened 2023-09-11
SUP-1354: Cluster Queues integration

james2791:
feat(Cluster Queues): [Adds support for interacting with Cluster queue REST API functionality]

PR checklist:
- [x] tests added 
- [x] examples of each service action (List, Get etc) added to the relevant `examples/` folder (create if new)

### DIFF
--- a/buildkite/clusters.go
+++ b/buildkite/clusters.go
@@ -29,22 +29,22 @@ type ClusterUpdate struct {
 }
 
 type Cluster struct {
-	ID              *string      `json:"id,omitempty" yaml:"id,omitempty"`
-	GraphQLID       *string      `json:"graphql_id,omitempty" yaml:"graphql_id,omitempty"`
-	DefaultQueueID  *string      `json:"default_queue_id,omitempty" yaml:"default_queue_id,omitempty"`
-	Name            *string      `json:"name,omitempty" yaml:"name,omitempty"`
-	Description     *string      `json:"description,omitempty" yaml:"description,omitempty"`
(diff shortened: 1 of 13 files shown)


buildkite/clusters.go
```diff
@@ -29,22 +29,22 @@ type ClusterUpdate struct {
...
+	URL             *string      `json:"url,omitempty" yaml:"url,omitempty"`
+	WebURL          *string      `json:"web_url,omitempty" yaml:"web_url,omitempty"`
+	QueuesURL       *string      `json:"queues_url,omitempty" yaml:"queues_url,omitempty"`
+	DefaultQueueURL *string      `json:"default_queue_url,omitempty" yaml:"default_queue_url,omitempty"`
+	CreatedAt       *Timestamp   `json:"created_at,omitempty" yaml:"created_at,omitempty"`
+	CreatedBy       *ClusterUser `json:"created_by,omitempty" yaml:"created_by,omitempty"`
 }
 
-type ClusterCreator struct {
+type ClusterUser struct {
```
mcncl, line 47: This change seems like it's less idiomatic than the original? Isn't everyone a cluster user? A cluster creator feels as though it's more indicative of the actual _thing_.

mcncl (changes requested):

buildkite/clusters.go
james2791: Hmm good point - I was honestly on the fence, though the rename was more that the same struct instances is used for the ClusterQueue `CreatedBy` attribute too, but I can see the naming case there (ie not implementing a duplicate/same attributes `ClusterQueueCreator` struct)

PR #173, opened 2024-01-16
Expose additional webhook fields on builds and jobs

mstifflin:
This is useful when deserializing json webhooks that have this data.


james2791:
Thanks for contributing this @mstifflin 🙂 Looks good - and shall be looped into the next release

james2791 (approved):

Issue #170, opened 2024-01-08
Can't construct webhook objects using literals


mattdee123:
**Describe the bug**
In order to write a test for a webhook handler, I would like to create sample webhook events, but cannot do so with a struct literal expression.

**To Reproduce**
Try to construct the following object:
```
&buildkite.JobFinishedEvent{Pipeline: &buildkite.Pipeline{}}
```

This gives a compiler error. However,

```
event := &buildkite.JobFinishedEvent{}
event.Pipeline = &buildkite.Pipeline{}
```

**Expected behavior**
It would be nice if the above pattern, or something similar, could be used to construct test objects.

**Additional context**

This happens because the type is 
```
type JobFinishedEvent struct {
	jobEvent
}
```
and the embedded unexported field means we can't use a struct literal to create it, which is slightly annoying when writing tests.

jradtilbrook:
Hi @mattdee123 thanks for raising this usability improvement. I'm wondering if we should do away with the embedding and just us a type alias. I'm unsure the reasoning why embedding was used in the first place.

We'll investigate some options here, but I don't think this is a bug and there is a simple workaround for it at the moment

mcncl:
Doing some housekeeping, this was completed in https://github.com/buildkite/go-buildkite/pull/175

PR #175, opened 2024-01-30
SUP-1681: Webhook event type definition for struct literal creation

james2791:
Feat (Webhook event types): Webhook event type definitions for allowing creation via struct literals

PR checklist:
- [ ] tests added -> (All pipeline template interaction with the REST API) -> *Tests consistent*
- [ ] examples of each service action (List, Get etc) added to the relevant `examples/` folder -> **No services altered**

In previous version of Go-Buildkite; the `type`s of agent, build, job webhook events were created as structs without inner field exporting; meaning that instances of these types couldn't be created in a struct literal manner. A instance of the base type was required _first_ before adding its sub-fields:

```
jobFinishedEvent := buildkite.JobFinishedEvent{}
jobFinishedEvent.Pipeline = &buildkite.Pipeline{
	Name: buildkite.String("Android App"),
	... additional fields ...
}
```

This PR exports the inner types so that instances can be struct literal created while also keeping the above consuming model above in place - for example:

```
jobFinishedEvent := buildkite.JobFinishedEvent{
	JobEvent: buildkite.JobEvent{
		Pipeline: &buildkite.Pipeline{
			Name: buildkite.String("Android App"),
		},
		Build: ...
		Job: ...
	},
}
```

Addresses #170

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,5 +1,8 @@
 ## Unreleased
 * Expose retry_source and retry_type on jobs [#171](https://github.com/buildkite/go-buildkite/pull/171) ([drcapulet](https://github.com/drcapulet))
+* Expose additional webhook fields on builds and jobs [#173](https://github.com/buildkite/go-buildkite/pull/173) ([mstifflin](https://github.com/mstifflin))
+* SUP-1697 Add username to Author struct [#174](https://github.com/buildkite/go-buildkite/pull/174) ([lizrabuya](https://github.com/lizrabuya))
+* SUP-1681: Webhook event type definition for struct literal creation [#175](https://github.com/buildkite/go-buildkite/pull/175) ([james2791](https://github.com/james2791))
 
 ## [v3.10.0](https://github.com/buildkite/go-buildkite/compare/v3.9.0...v3.10.0) (2023-11-15)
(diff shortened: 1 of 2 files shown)


mcncl:
This change also means that the aliased types cannot have their own fields, they must use all and only the fields of their "event" types.

mcncl:
Would require a `major` bump due to lack of backwards compatibility.

jradtilbrook (commented):
What if we just exported the inner fields so they could be accessed? I don't know if keeping them private is that useful

james2791:
👆

How about typed exporting - similar to what I had implemented [first](https://github.com/buildkite/go-buildkite/pull/175/commits/f1a978bd0fc8b2eff38fbc7988f7e1d2df39602a)? It would keep functionality, Im not too sure on why they were privatised in the beginning (minus exporting the base `AgentEvent` etc)

jradtilbrook:
Yeah I think that seems okay. And that should be backwards compatible as well right?

james2791:
Correct - same process in creating these struct instances as before; export just gives the chance to create them with the inner fields. Thankfully its not a drastic type change that would need to constitute a `x.0.0` release 👍

jradtilbrook (approved):

PR #178, opened 2024-02-26
CODEOWNERS update

james2791:
Adjusted CODEOWNERS with current active contributors.

Still happy to review things, just on a best effort, non-notification basis! 💚


mcncl (approved):
💥

PR #183, opened 2024-05-15
First cut of fetch test plan endpoint.

wooly:
Here's a (probably) very rough cut of a new API endpoint for fetching TestPlans for test splitting. This endpoint is currently being hit by the [test-splitter](https://github.com/buildkite/test-splitter) itself, but I think it makes sense for the go bindings to grow support as well.

cc @nprizal and @amybiyuliu

### DIFF
--- a/buildkite/buildkite.go
+++ b/buildkite/buildkite.go
@@ -57,7 +57,7 @@ type Client struct {
 	User              *UserService
 	Teams             *TeamsService
 	Tests             *TestsService
-	TestPlans					*TestPlansService
+	TestPlans         *TestPlansService
 	TestRuns          *TestRunsService
 	TestSuites        *TestSuitesService
 }
--- a/buildkite/buildkite.go
+++ b/buildkite/buildkite.go
@@ -57,6 +57,7 @@ type Client struct {
 	User              *UserService
 	Teams             *TeamsService
 	Tests             *TestsService
+	TestPlans					*TestPlansService
 	TestRuns          *TestRunsService
 	TestSuites        *TestSuitesService
 }
@@ -97,6 +98,7 @@ func NewClient(httpClient *http.Client) *Client {
 	c.User = &UserService{c}
 	c.Teams = &TeamsService{c}
 	c.Tests = &TestsService{c}
+	c.TestPlans = &TestPlansService{c}
 	c.TestRuns = &TestRunsService{c}
 	c.TestSuites = &TestSuitesService{c}
 
--- a/buildkite/test_plans.go
+++ b/buildkite/test_plans.go
@@ -19,7 +19,6 @@ type Tests struct {
 type Task struct {
 	NodeNumber *int   `json:"node_number,omitempty" yaml:"node_number,omitempty"`
 	Tests      *Tests `json:"tests,omitempty" yaml:"tests,omitempty"`
-
 }
 
 type TestPlan struct {
--- a/buildkite/test_plans.go
+++ b/buildkite/test_plans.go
@@ -29,7 +29,6 @@ type TestPlan struct {
 
 func (tps *TestPlansService) Get(org, slug string, identifier string) (*TestPlan, *Response, error) {
 	u := fmt.Sprintf("v2/analytics/organizations/%s/suites/%s/test_plan?identifier=%s", org, slug, identifier)
-	fmt.Printf(u)
 
 	req, err := tps.client.NewRequest("GET", u, nil)
 
--- a/buildkite/test_plans.go
+++ b/buildkite/test_plans.go
@@ -1,0 +1,50 @@
+package buildkite
+
+import "fmt"
+
+type TestPlansService struct {
+	client *Client
+}
+
+type Case struct {
+	Path              *string `json:"path,omitempty" yaml:"path,omitempty"`
+	EstimatedDuration *int `json:"estimated_duration,omitempty" yaml:"estimated_duration,omitempty"`
+}
+
+type Tests struct {
+	Format *string `json:"format,omitempty" yaml:"format,omitempty"`
+	Cases  []Case  `json:"cases,omitempty" yaml:"cases,omitempty"`
+}
+
+type Task struct {
+	NodeNumber *int   `json:"node_number,omitempty" yaml:"node_number,omitempty"`
+	Tests      *Tests `json:"tests,omitempty" yaml:"tests,omitempty"`
+
+}
+
+type TestPlan struct {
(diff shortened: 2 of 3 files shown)


buildkite/test_plans.go
```diff
@@ -0,0 +1,50 @@
...
+
+type Case struct {
+	Path              *string `json:"path,omitempty" yaml:"path,omitempty"`
+	EstimatedDuration *int `json:"estimated_duration,omitempty" yaml:"estimated_duration,omitempty"`
+}
+
+type Tests struct {
+	Format *string `json:"format,omitempty" yaml:"format,omitempty"`
+	Cases  []Case  `json:"cases,omitempty" yaml:"cases,omitempty"`
+}
```
mcncl, line 17: Everything else has a singular, would we have a Test struct or is that just overkill? Is there ever a time when we would need a singular test? Or is `test == case`?

buildkite/test_plans.go
```diff
@@ -0,0 +1,50 @@
...
+
+type Task struct {
+	NodeNumber *int   `json:"node_number,omitempty" yaml:"node_number,omitempty"`
+	Tests      *Tests `json:"tests,omitempty" yaml:"tests,omitempty"`
+
```
mcncl, line 22: This empty line might get picked up by a linter.

buildkite/test_plans_test.go
```diff
@@ -0,0 +1,98 @@
...
+	"reflect"
+	"testing"
+)
+
+func TestTestPlansService_Get(t *testing.T) {
+	setup()
+	defer teardown()
+
+	mux.HandleFunc("/v2/analytics/organizations/my-great-org/suites/suite-1/test_plan", func(w http.ResponseWriter, r *http.Request) {
+		testMethod(t, r, "GET")
```
mcncl, line 15: `testMethod` being in a different file feels hard to eye trace. Not a you change, just noticed it.

wooly: Yeah it threw me a little at first. But then I guess `mux` is also in a different file so I assumed things were just coming from somewhere else!

buildkite/test_plans.go
wooly: :+1:

buildkite/test_plans.go
wooly: Roughly speaking `tests == case` is true. At the moment, an individual case is a file containing tests that is run by the test splitter. The `format` is set to `file` to denote that all cases within the cases array are file-level.

buildkite/test_plans.go
wooly: Fixed by https://github.com/buildkite/go-buildkite/pull/183/commits/77f55d7de1092e71f3c438f2c0b1433beb3f49f0

buildkite/buildkite.go
```diff
@@ -57,6 +57,7 @@ type Client struct {
 	User              *UserService
 	Teams             *TeamsService
 	Tests             *TestsService
+	TestPlans					*TestPlansService
```
nprizal, line 60: [The nail that sticks out gets hammered down](https://en.wiktionary.org/wiki/the_nail_that_sticks_out_gets_hammered_down)

buildkite/test_plans.go
```diff
@@ -0,0 +1,49 @@
+package buildkite
+
+import "fmt"
+
+type TestPlansService struct {
+	client *Client
+}
+
+type Case struct {
+	Path              *string `json:"path,omitempty" yaml:"path,omitempty"`
```
nprizal, line 10: I don't see a reason this needs to be a pointer. The path will always be in the response.

buildkite/test_plans.go
```diff
@@ -0,0 +1,49 @@
...
+	EstimatedDuration *int `json:"estimated_duration,omitempty" yaml:"estimated_duration,omitempty"`
+}
+
+type Tests struct {
+	Format *string `json:"format,omitempty" yaml:"format,omitempty"`
```
nprizal, line 15: Same as above

buildkite/test_plans.go
```diff
@@ -0,0 +1,49 @@
...
+	Cases  []Case  `json:"cases,omitempty" yaml:"cases,omitempty"`
+}
+
+type Task struct {
+	NodeNumber *int   `json:"node_number,omitempty" yaml:"node_number,omitempty"`
+	Tests      *Tests `json:"tests,omitempty" yaml:"tests,omitempty"`
```
nprizal, line 21: same as above

buildkite/test_plans.go
```diff
@@ -0,0 +1,49 @@
...
+}
+
+type TestPlan struct {
+	Identifier  *string `json:"identifier,omitempty" yaml:"identifier,omitempty"`
+	Parallelism *int    `json:"parallelism,omitempty" yaml:"paralellism,omitempty"`
```
nprizal, line 26: Same as above

buildkite/test_plans.go
```diff
@@ -0,0 +1,49 @@
...
+	Tasks       []Task  `json:"tasks,omitempty" yaml:"tasks,omitempty"`
+}
+
+func (tps *TestPlansService) Get(org, slug string, identifier string) (*TestPlan, *Response, error) {
+	u := fmt.Sprintf("v2/analytics/organizations/%s/suites/%s/test_plan?identifier=%s", org, slug, identifier)
+	fmt.Printf(u)
```
nprizal, line 32:
```suggestion
```
Is this intentional? The request URL will be printed when the debug mode is enabled.

wooly: Nope, will snip!

wooly:
Does this look like it's ready to go?

buildkite/test_plans.go
```diff
@@ -0,0 +1,48 @@
...
+
+type Case struct {
+	Path              *string `json:"path,omitempty" yaml:"path,omitempty"`
+	EstimatedDuration *int `json:"estimated_duration,omitempty" yaml:"estimated_duration,omitempty"`
+}
+
+type Tests struct {
+	Format *string `json:"format,omitempty" yaml:"format,omitempty"`
+	Cases  []Case  `json:"cases,omitempty" yaml:"cases,omitempty"`
+}
```
nprizal, line 17: We've changed the schema. `Tests` is now just an array of `Case`.

wooly: Ah yeah so we did, will need to update!

wooly:
Closing this as not a priority for us right now.

PR #185, opened 2024-07-18
Deprecate authenticating roundtrippers, replace with new client constructor

moskyb:
Using RoundTrip to modify requests is [discouraged](https://pkg.go.dev/net/http#RoundTripper) by the go authors, and given that we control the full lifecycle of the request anyway, we can create a nicer interface for users of this package.

To wit, this commit deprecates the `buildkite.New` function, as well as the authenticating transports that it's used with - creating a client would usually look like:

```go
	config, err := buildkite.NewTokenConfig(apiToken, debug)
	if err != nil {
		return fmt.Errorf("client config failed: %w", err)
	}

	client := buildkite.NewClient(config.Client())
```

This commit adds a new constructor method for clients, `NewOpts`. It accepts options to configure both of our supported auth schemes:
```go
client := buildkite.NewOpts(buildkite.WithTokenAuth(token))
client = buildkite.NewOpts(buildkite.WithBasicAuth("shirley_dander", "hunter2"))
```

`NewOpts` also accepts options for configuring the http client it uses, the base URL it uses for requests, and the user agent it reports.

### DIFF
--- a/buildkite/access_tokens_test.go
+++ b/buildkite/access_tokens_test.go
@@ -8,7 +8,7 @@ import (
 )
 
 func TestAccessTokensService_Get(t *testing.T) {
-	setup()
+	setup(t)
 	defer teardown()
 
 	mux.HandleFunc("/v2/access-token", func(w http.ResponseWriter, r *http.Request) {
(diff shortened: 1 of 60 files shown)


james2791 (approved):
🚀 :shipit:

PR #187, opened 2024-08-08
Test cleanups

moskyb:
**This PR:** Makes a grab-bag of improvements to the testing of this library. Most notably, it updates the buildkite pipeline to include some basic linting, and splits the existing step up into two separate steps, but it also:
- Fixes some invalid struct tags
- Parallelises the tests (and makes each test case independent from one another by making them not share global vars)
- Updates to go 1.22
- Makes tests fail when an unexpected HTTP request is received by the test server
- Uses `t.Cleanup()` instead of `defer` in tests

**This PR is best reviewed commit-by-commit**

### DIFF
--- a/.buildkite/pipeline.yml
+++ b/.buildkite/pipeline.yml
@@ -1,6 +1,26 @@
 steps:
-  - name: ":golang:"
-    command: make
+  - name: ":go::robot_face: Lint"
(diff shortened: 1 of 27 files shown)


james2791 (approved):
🚀 :shipit:

PR #188, opened 2024-08-08
Add context parameters to all methods provided by this library

### LINKED ISSUE #98: support using go context
In Golang, propagating context through the stack is very important:

1. Let you retain a certain metadata
2. Let you handle graceful shutdown logic
3. Let you implement cancel/timeout

Current library has not support for making http request with a context.

We could introduce `RequestOptionFunc` to help handle this (see https://github.com/xanzy/go-gitlab/blob/7d79a18484610e0d7b178392f020c4a8dfb419c6/applications.go#L75-L91 for reference)

```
// RequestOptionFunc can be passed to all API requests to customize the API request.
type RequestOptionFunc func(*retryablehttp.Request) error[Sander van Harmelen, 2 years ago: • Introduce client options](https://sourcegraph.com/github.com/xanzy/go-gitlab/-/commit/443faa191281c0e6dd2d6df9c29c2d1caa515d4d)

// WithContext runs the request with the provided context
func WithContext(ctx context.Context) RequestOptionFunc {
	return func(req *retryablehttp.Request) error {
		*req = *req.WithContext(ctx)
		return nil
	}
}
```

The modify existing interfaces to 

```
func (*obj) ListObj(lo ListObjOptions, options ...RequestOptionFunc) ([]*Obj, *Response, error) {}
```

This would help it so that if parent application decided to shutdown half way through, we can cancel the in-flight request early (before the server even response)

moskyb:
**This PR:** Adds `context.Context` parameters to all methods exported by this library that make API calls to the Buildkite API. **This is a breaking change to the API and will necessitate a v4 release.**

Contexts are useful in this case because they allow users to customise a lot of the behaviour of the requests, most notably by adding timeouts and allowing users to cancel requests - eg, a user would be able to do something like

```Go
ctx, cancel := context.WithDeadline(context.Background(), 5*time.Second)
client.Builds.List(ctx, /* ... */)
```
which would time the request out after five seconds. they'd also be able to use the `cancel` function to cancel the request from another goroutine, for whatever reason they choose.

Closes #98

### DIFF
--- a/buildkite/access_tokens.go
+++ b/buildkite/access_tokens.go
@@ -1,6 +1,9 @@
 package buildkite
 
-import "fmt"
+import (
+	"context"
+	"fmt"
+)
 
 type AccessTokensService struct {
 	client *Client
@@ -14,13 +17,13 @@ type AccessToken struct {
 // Get gets the current token which was used to authenticate the request
 //
 // buildkite API docs: https://buildkite.com/docs/rest-api/access-token
-func (ats *AccessTokensService) Get() (*AccessToken, *Response, error) {
+func (ats *AccessTokensService) Get(ctx context.Context) (*AccessToken, *Response, error) {
 
 	var u string
 
 	u = fmt.Sprintf("v2/access-token")
 
-	req, err := ats.client.NewRequest("GET", u, nil)
+	req, err := ats.client.NewRequest(ctx, "GET", u, nil)
 	if err != nil {
 		return nil, nil, err
 	}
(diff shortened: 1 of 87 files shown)


jradtilbrook:
@moskyb we have a few other plans for a v4 breaking change in 2025 so would like to wait for then. You can find some ideas in Linear for this repo. Would be great to get your input on it as well!

moskyb:
sounds great! i'll stop drive-by-ing now, promise :P

DrJosh9000 (approved):
NICE

mcncl:
Love this. Not had a chance to review it but definitely in favour of this functionality and agree that it would necessitate a major version bump.

mcncl (approved):

PR #192, opened 2024-09-01
Updated README with NewOpts client/kingpin example

james2791:
Release [v3.12.0](https://github.com/buildkite/go-buildkite/releases/tag/v3.12.0) introduced a new client configuration constructor via `NewOpts`. The first example on the README was left with the older way of constructing a new client via `NewTokenConfig` - which this adjusts with `NewOpts` and a simplified example with kingpin.

### DIFF
--- a/README.md
+++ b/README.md
@@ -19,7 +19,7 @@ import (
 )
 
 var (
-	apiToken = kingpin.Flag("token", "API token").Required().String()
+    apiToken = kingpin.Flag("token", "API token").Required().String()
     org = kingpin.Flag("org", "Organization slug").Required().String()
 )
 
--- a/README.md
+++ b/README.md
@@ -15,19 +15,21 @@ Simple shortened example for listing all pipelines:
 ```go
 import (
     "github.com/buildkite/go-buildkite/v3/buildkite"
-)
-...
 
-config, err := buildkite.NewTokenConfig(*apiToken, false)
+    "gopkg.in/alecthomas/kingpin.v2"
+)
+
+var (
+	apiToken = kingpin.Flag("token", "API token").Required().String()
+)
+
+client, err := buildkite.NewOpts(buildkite.WithTokenAuth(*apiToken))
 
 if err != nil {
-	log.Fatalf("client config failed: %s", err)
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -15,19 +15,21 @@ Simple shortened example for listing all pipelines:
...
+var (
+    apiToken = kingpin.Flag("token", "API token").Required().String()
+    org = kingpin.Flag("org", "Organization slug").Required().String()
+)
+
+client, err := buildkite.NewOpts(buildkite.WithTokenAuth(*apiToken))
 
 if err != nil {
-	log.Fatalf("client config failed: %s", err)
+	log.Fatalf("New client configuration failed: %s", err)
```
mcncl, line 29: It's likely that this would be wrapped, so best to keep it lower case.

I also don't think we need "new".

README.md
```diff
@@ -15,19 +15,21 @@ Simple shortened example for listing all pipelines:
 ```go
 import (
     "github.com/buildkite/go-buildkite/v3/buildkite"
+    "gopkg.in/alecthomas/kingpin.v2"
 )
-...
 
-config, err := buildkite.NewTokenConfig(*apiToken, false)
+var (
```
mcncl, line 21: Did we add this in for a reason? I don't see them in the original README. This isn't a requested change

mcncl (changes requested):

README.md
james2791: Reverted back 👍

README.md
james2791: @mcncl more to have the READMEs' example in-line with our actual examples in that dir, though more the very shorthand example. We were also referencing `*org` and `*apiToken` without their actual assignment before so its more a concrete example just in case it was copy-pasted 👍

mcncl: Yeah, I think the `...` is intended to be a _other code here_ type thing.

mcncl (approved):

README.md
james2791: Oh yeah thats fair - I guess now its more concrete than anything :)

PR #194, opened 2024-10-09
[WIP] Package presigned upload

moskyb:


buildkite/package_uploads.go
```diff
@@ -0,0 +1,216 @@
...
+	}
+
+	p.OrganizationSlug = organizationSlug
+	p.RegistrySlug = registrySlug
+
+	return p, resp, err
+}
+
+// Perform performs uploads the package file referred to by `file` to the presigned upload URL.
+// It does not create the package in the registry, only uploads the file to S3. The returned string is the URL of the
```
moskyb, line 134:
```suggestion
// It does not create the package in the registry, only uploads the file to the package host. The returned string is the URL of the
```

buildkite/package_uploads.go
```diff
@@ -0,0 +1,216 @@
...
+
+	req.Header.Set("Content-Type", s.ContentType)
+
+	// Don't set the Content-Length header here, you fool, you absolute buffoon
+	// When passed an io.Reader, http.NewRequestWithContext will not set the Content-Length header, and will instead
+	// stream the request body. This _would_ be exactly what we want, except that S3 uploads infuriatingly require a
+	// Content-Length header. So we have to calculate the length of the request body ourselves and set it manually on the
+	// request. Adding:
+	// 	req.Header.Set("Content-Length", fmt.Sprintf("%d", s.Len()))
+	// is not sufficient, as the Content-Length header is stripped by the http client when the request body is an io.Reader.
```
moskyb, line 167: i'm not particularly tied to this comment, it was just something i learned along the way. happy to get rid of it or keep it, if anyone has opinions
```suggestion
```

buildkite/package_uploads.go
```diff
@@ -0,0 +1,216 @@
...
+	req.ContentLength = s.Len()
+
+	resp, err := http.DefaultClient.Do(req)
```
moskyb, line 170: should we be using an internal HTTP client from the buildkite client?

buildkite/package_uploads.go
```diff
@@ -0,0 +1,216 @@
...
+	p.OrganizationSlug = organizationSlug
+	p.RegistrySlug = registrySlug
+
+	return p, resp, err
+}
+
+// Perform performs uploads the package file referred to by `file` to the presigned upload URL.
+// It does not create the package in the registry, only uploads the file to S3. The returned string is the URL of the
+// uploaded file in S3, which can then be passed to [Finalize] to create the package in the registry.
+func (ppu PackagePresignedUpload) Perform(ctx context.Context, ps *PackagesService, file *os.File) (string, error) {
```
moskyb, line 136: this method signature is a little odd compared to the other ones in this library - the `PackagePresignedUpload` is the receiver, and the `*PackagesService` is an arg. should it be the other way around?

my logic here was that the `PackagePresignedUpload` was the key object, and thus "deserved" to be the receiver, but i'm happy to change this to the more standard `*PackagesService` receiver.

One benefit to this approach is that method names are cleaner
```Go
ppu.Perform(ctx, ps, f)
// vs
ps.PerformPresignedUpload(ctx, ppu, f)
```

moskyb: that said though, most of the "complexity" of this API is elided to the user - 99.999% of people are going to call `ps.Create`, which handles all of this from them

buildkite/package_uploads.go
```diff
@@ -0,0 +1,216 @@
...
+	if err != nil {
+		return Package{}, nil, fmt.Errorf("performing presigned upload: %v", err)
+	}
+
+	p, resp, err := ppu.Finalize(ctx, ps, s3URL)
+	if err != nil {
+		return Package{}, nil, fmt.Errorf("finalizing package: %v", err)
+	}
+
+	return p, resp, nil
```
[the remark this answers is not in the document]
DrJosh9000: What is the response return used for? Looks like it is ignored when it is called below. Perhaps delete it?

If `Create` is implementing a "workflow" (i.e. sequence of HTTP requests and responses) then is expecting the caller to do something with the `Response` as well as the `Packge` leaking the abstraction a bit?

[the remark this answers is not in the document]
moskyb: i _think_ for most use cases, the response returned can be used for more inspection into errors/deeper introspection of HTTP.... stuff. to be honest, i only added it to this method because every other public method in this module returns a tuple of `[T any](T, *Response, error)`, and it felt like it'd be weird to have one method that doesn't do that?'

`Create` is indeed implementing a workflow - you might be right about making it just return `(Package, error)`

tommeier:
https://linear.app/buildkite/issue/PKG-7648/[cli]-bk-package-push-direct-s3-upload

buildkite/package_uploads.go
```diff
@@ -0,0 +1,216 @@
...
+		return Package{}, nil, fmt.Errorf("writing package to tempfile: %v", err)
+	}
+	defer func() {
+		packageTempFile.Close()
+		os.Remove(packageTempFile.Name())
+	}()
+
+	s := bkmultipart.NewStreamer()
+	s.WriteFile(fileFormKey, packageTempFile, filename)
+
```
moskyb, line 41: do i need this? it might be handled by `PresignedPackageUpload` internally

buildkite/package_uploads.go
james2791: For total transparency the first approach with `ppu`/`PackagePresignedUpload` receiver and the simpler method names gets my vote. Its more effectively "knowing" that any method on said receiver belongs to a pre-signed upload; whereas `ps` generally could have a whole lot of methods attached which is common across all the other services!

buildkite/package_uploads.go
moskyb: i do not!

buildkite/package_uploads.go
moskyb: good enough for me!

PriyaSudip (approved):

PR #195, opened 2024-10-14
Package improvements grab bag

moskyb:
**This PR:** Adds a couple of features/fixes to the package registry functions in this module, as well as a change to HTTP testing that'll allow for better multi-request testing.

The changes:
- Previously, we were buffering package files in memory twice(!!) before sending them off to buildkite - once in the multipart body, then again as part of sending the request off. This PR gets rid of the first buffer in 24ec51eb9dfa81220f63349c5ebf7f6f31e78c40, and the second in e080be79093c71fc00c037c4759361ba27b6ad4b. Files are now streamed to buildkite without being stored in memory.

  Part of this work was adapting/stealing the multipart streamer [from the agent](https://github.com/buildkite/agent/blob/main/internal/artifact/bk_uploader.go#L299-L370) and refining it a wee bit - this is all tested, but i'd love some extra eyes on it!
  
- We've recently added a bunch more fields to package registries in the Buildkite REST API (thanks @ensvo!!). This PR updates structs reflecting registries to include those new fields
  -  Note: we only return the OIDC policy as a string, as it's a non-consistent structure and tricky to parse idiomatically in go. We do, however, offer a type, `buildkite.OIDCPolicy` that allows for idiomatic construction of OIDC policies in raw go.
- I've added a feature to our HTTP testing in this module so that the test server records its calls - this makes it easier for us to assert that a certain number of calls were made, or which calls were made in which order.

### DIFF
--- a/buildkite/packages.go
+++ b/buildkite/packages.go
@@ -61,14 +61,16 @@ func (ps *PackagesService) Create(ctx context.Context, organizationSlug, registr
 	if err != nil {
 		return Package{}, nil, fmt.Errorf("writing package to tempfile: %v", err)
 	}
-	defer os.Remove(packageTempFile.Name())
-	defer packageTempFile.Close()
+	defer func() {
+		packageTempFile.Close()
+		os.Remove(packageTempFile.Name())
+	}()
 
 	s := bkmultipart.NewStreamer()
 	s.WriteFile(fileFormKey, packageTempFile, filename)
 
 	url := fmt.Sprintf("v2/packages/organizations/%s/registries/%s/packages", organizationSlug, registrySlug)
-	req, err := ps.client.NewRequest(ctx, "POST", url, s)
+	req, err := ps.client.NewRequest(ctx, "POST", url, s.Reader())
 	if err != nil {
 		return Package{}, nil, fmt.Errorf("creating POST package request: %v", err)
 	}
@@ -85,9 +87,8 @@ func (ps *PackagesService) Create(ctx context.Context, organizationSlug, registr
 	return p, resp, err
 }
 
-// normalizeToFile takes and io.Reader (which might itself already be a file, but could be a stream or other source) and
+// normalizeToFile takes an io.Reader (which might itself already be a file, but could be a stream or other source) and
 // writes it to a temporary file, returning the file handle.
-// The file is written to a temporary directory, and then renamed to the desired filename.
 // We do this normalization to ensure that we can accurately calculate the Content-Length of the request body, which is
 // required by S3. We write to disk (instead of buffering in memory) to avoid memory exhaustion for large files.
 func normalizeToFile(r io.Reader, filename string) (*os.File, error) {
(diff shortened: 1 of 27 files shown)


internal/bkmultipart/streamer_test.go
```diff
@@ -0,0 +1,114 @@
...
+func TestEncodingFile(t *testing.T) {
+	t.Parallel()
+
+	tempFile, err := os.CreateTemp("", "buildkite-go-sdk-test")
+	if err != nil {
+		t.Fatalf("os.CreateTemp() = %v, want nil", err)
+	}
+
+	t.Cleanup(func() { os.Remove(tempFile.Name()) })
+	t.Cleanup(func() { tempFile.Close() })
```
DrJosh9000, line 47: Since you need to wrap the calls in a func anyway, you could do this (making the order of operations clearer):
```suggestion
	t.Cleanup(func() {
		tempFile.Close()
		os.Remove(tempFile.Name())
	})
```

DrJosh9000 (approved):
Nice, we should reverse-steal the improvements back into the agent...

(something something monorepo)

buildkite/packages.go
```diff
@@ -85,3 +84,35 @@ func (ps *PackagesService) Create(ctx context.Context, organizationSlug, registr
...
+	if err != nil {
+		return nil, fmt.Errorf("creating temporary file: %v", err)
+	}
+
+	_, err = io.Copy(f, r)
+	if err != nil {
+		return nil, fmt.Errorf("writing to temporary file: %v", err)
+	}
+
+	_, err = f.Seek(0, 0)
```
wolfeidau, line 105: I prefer to using the constant for the second option `f.Seek(0, io.SeekStart)` as it makes it more approachable.

internal/bkmultipart/streamer_test.go
```diff
@@ -0,0 +1,114 @@
...
+
+	if _, err := tempFile.WriteString("hello world"); err != nil {
+		t.Fatalf(`tempFile.WriteString("hello world") = %v, want nil`, err)
+	}
+
+	if _, err := tempFile.Seek(0, 0); err != nil {
```
wolfeidau, line 53: I prefer to using the constant for the second option `f.Seek(0, io.SeekStart)` as it makes it more approachable.

wolfeidau (changes requested):
Just some small niggles, otherwise it seems like a good tidy up.

Multi part is a 🗑️ 🔥 I mostly trust other peoples packages as you have mentioned, so many edge cases.

james2791 (changes requested):
Tiny quick one in addition to @wolfeidau / @DrJosh9000's 👍

internal/bkmultipart/streamer.go
```diff
@@ -0,0 +1,92 @@
...
+	s := &Streamer{bodyBuffer: &bytes.Buffer{}}
+
+	s.bodyWriter = multipart.NewWriter(s.bodyBuffer)
+	boundary := s.bodyWriter.Boundary()
+	s.ContentType = "multipart/form-data; boundary=" + boundary
+
+	closeBoundary := fmt.Sprintf("\r\n--%s--\r\n", boundary)
+	s.closeBuffer = bytes.NewBufferString(closeBoundary)
+
+	// if a file is added, the reader will be set up to read from the body, the file and the closer
```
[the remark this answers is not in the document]
moskyb: yeah... i might refactor this so that it has an internal slice of `io.Reader`s that it appends to, and have it have a method like 
```Go
func (s *Streamer) Reader io.Reader {
  r := io.MultiReader(s.readers...)
  r = io.MultiReader(r, closeBuffer)
}
```

which would remove the necessity to only call `WriteFile` once, and allow the files to be put anywhere in the body (though in our case they always need to be at the end)

internal/bkmultipart/streamer_test.go
moskyb: ooh yes, good call

buildkite/packages.go
moskyb: yep, good call, will add

wolfeidau (approved):

james2791 (approved):
Will need codeownership ✅ 👍

mcncl (approved):
Sure there are already 3 approvals. What could go wrong

PR #196, opened 2024-10-14
Remove YAML bindings for exported data types

moskyb:
Buildkite does not output any API responses in YAML, nor (in general) does it accept them. There are exceptions to the acceptance side of that, but within this library, they'll either be passed as JSON (a subset of YAML), or passed as raw text.

This removes the ability for external users to make calls like
```Go
bytes, err := yaml.Marshal(Agent{/* ... */})
```
(or more accurately; it'll remove the ability for those calls to work as they currently do - names of keys will not be properly cased and yaml library semantics like `inline` and `omitempty` will not function)

**This is a breaking change.**

### DIFF
--- a/buildkite/access_tokens.go
+++ b/buildkite/access_tokens.go
@@ -10,8 +10,8 @@ type AccessTokensService struct {
 }
 
 type AccessToken struct {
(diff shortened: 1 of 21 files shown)


mcncl (approved):
Will be a major version bump but looks good.

PR #197, opened 2024-10-15
Remove pointer arguments where they're not needed

moskyb:
**This PR:** Is a bit of a doozy!

When initially creating this library, we got super trigger happy with arguments that are pointers, to the point of making most arguments pointers to structs, and most of the members of those structs pointers to scalar types.

in 99% of cases, most of these things don't need to be pointers, so i've done a refactor with the philosophy of:
- Unless something _must_ be a pointer - that is, there's a meaningful semantic difference between `nil` and its empty value - don't use pointers
- If structs represent data only - they don't have any methods defined on them - return them as values, not pointers
- If structs as arguments don't need to be modified, and there's no no meaningful semantic difference between nil and zero values, don't use pointers to structs as arguments
  - Notably, there's one strain of argument that we _keep_ as pointers, and that's a variety of List options, eg [here](https://github.com/buildkite/go-buildkite/pull/197/files#diff-f8d935955c893b7714976c2064cf494781474b6009a8763b0384ed09b04a4ed8L58). In this case i _do_ think there's a meaningful semantic difference between zero and nil values, but i'm happy to argue about this
- In general, don't expose any unmarshalling semantics to users, again unless necessary. There was a pattern in this repo previously where you would have something like:
```go
fooUpdate := buildkite.FooUpdate{/* set some fields to update /*}
resp, err := bkclient.FooService.Update(ctx, &fooUpdate)
```

where `fooUpdate` would be the fields to update, but then the result of the update call would also be unmarshalled into `fooUpdate`, rather than returned. this is an issue because encoding/json behaves surprisingly ([#27172](https://github.com/golang/go/issues/27172), [#31924](https://github.com/golang/go/issues/31924), [#26946](https://github.com/golang/go/issues/26946), [#21092](https://github.com/golang/go/issues/21092)) when unmarshalling into non-empty structs, and is also just kind of a weird user interface IMO. this behaviour is replaced with something like
```go
fooUpdate := buildkite.FooUpdate{/* set some fields to update /*}
updated, resp, err := bkclient.FooService.Update(ctx, fooUpdate)
```
where `updated` is a full `buildkite.Foo` struct. See b116725545b9df7b7d5b24f7f61e33234cec202d for more info.

### DIFF
--- a/buildkite/access_tokens.go
+++ b/buildkite/access_tokens.go
@@ -10,29 +10,24 @@ type AccessTokensService struct {
 }
 
 type AccessToken struct {
-	UUID   *string   `json:"uuid,omitempty"`
-	Scopes *[]string `json:"scopes,omitempty"`
+	UUID   string   `json:"uuid,omitempty"`
+	Scopes []string `json:"scopes,omitempty"`
 }
 
 // Get gets the current token which was used to authenticate the request
 //
 // buildkite API docs: https://buildkite.com/docs/rest-api/access-token
-func (ats *AccessTokensService) Get(ctx context.Context) (*AccessToken, *Response, error) {
-
-	var u string
-
-	u = fmt.Sprintf("v2/access-token")
-
+func (ats *AccessTokensService) Get(ctx context.Context) (AccessToken, *Response, error) {
+	u := fmt.Sprintf("v2/access-token")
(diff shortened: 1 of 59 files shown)


mcncl (approved):
🫗

PR #199, opened 2024-10-17
GolangCI Lint

moskyb:
**This PR:** Adds golangci-lint, a stricter and more opinionated version of go vet, into this library, and fixes the issues it found in this codebase. Most of the issues were us not checking the error returns for function calls (mostly calls to unmarshal JSON), but there were a couple of other bits and bobs as well.


mcncl (approved):
Love `golangci-lint`. This with `gofumpt` is my go-to local error checking.

PR #201, opened 2024-10-17
Remove deprecated methods ahead of v4 release

moskyb:
The `buildkite.NewClient` method was deprecated in favour of the `buildkite.NewOpts` method in #185. This PR renames `NewOpts` to `NewClient` (with its old method signature), and keeps `NewOpts` as an alias for the new `NewClient`. This means that users that have switched to the `NewOpts` initializer can stay using it.

The old client initializer, with its authenticating roundtrippers, have been removed entirely.


buildkite.go
```diff
@@ -142,31 +136,9 @@ func NewOpts(opts ...clientOpt) (*Client, error) {
...
-
-		if basicAuth, ok := c.client.Transport.(*BasicAuthTransport); ok {
-			basicAuth.APIHost = baseURL.Host
-		}
-	}
-	return c
+// NewOpts is an alias for NewClient
+func NewOpts(opts ...clientOpt) (*Client, error) {
+	return NewClient(opts...)
 }
```
mcncl, line 142: Hear me out...

```go
func NewClientWithOpts(opts ...clientOpts) (*Client, error) {
    return NewOpts(opts)
}
```

A `NewClientWithOpts` which returns a NewOpts which in turn returns a NewClient.

mcncl (approved):

Issue #211, opened 2025-01-05
Missing API for https://buildkite.com/docs/apis/agent-api/metrics


AustinSchuh:
I'm trying to fetch the queue depth to figure out if I need to spin up more workers or not.  https://buildkite.com/docs/apis/agent-api/metrics looks like the right API for that.

mcncl:
@AustinSchuh just circling on this one; it isn't possible to add this to the tool. That API call is made by an agent and thus would need an agent token, not an API token.

PR #212, opened 2025-01-05
support metrics api

mcncl:
## Changes
- Adds in support for a metrics service
- Provides a `GET` method on that service
- Tests
- Added to the core Buildkite client

Addresses https://github.com/buildkite/go-buildkite/issues/211


mcncl:
Closing this as I have been helped to realise an API token will not work for this call.

PR #215, opened 2025-04-09
Upgrade Go to 1.24

moskyb:
This PR updates Go to v1.24, which also necessitates an upgrade to `golangci-lint` to v2

### DIFF
--- a/.buildkite/pipeline.yml
+++ b/.buildkite/pipeline.yml
@@ -4,7 +4,7 @@ steps:
     command: .buildkite/steps/lint.sh
     plugins:
       - docker#v5.9.0:
-          image: "golangci/golangci-lint:v1.61-alpine"
+          image: "golangci/golangci-lint:v2.0-alpine"
 
   - name: ":go::test_tube: Test"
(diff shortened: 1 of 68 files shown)


.buildkite/pipeline.yml
```diff
@@ -4,15 +4,15 @@ steps:
     command: .buildkite/steps/lint.sh
     plugins:
       - docker#v5.9.0:
-          image: "golangci/golangci-lint:v1.61-alpine"
+          image: "golangci/golangci-lint:v2.0-alpine"
```
mcncl, line 7: There has been a couple of patches released, so the current version is `2.0.2`

mcncl (approved):

.buildkite/pipeline.yml
moskyb: `v2.0-alpine` should always point to the latest patch version of 2.0 -- happy to change this to point to a specific patch though

mcncl: TIL sounds great

PR #217, opened 2025-05-15
Export ClientOpt type for programmatic option composition

prateek:
This change exports the ClientOpt type (previously clientOpt) to allow users to build client options programmatically outside the package.

- Changed clientOpt -> ClientOpt with proper documentation
- Updated all associated function signatures
- Added example demonstrating the pattern in example_test.go
- Added test to verify the functionality

This provides a more flexible way to configure clients by combining options conditionally without nested if/else blocks.

### DIFF
--- a/buildkite.go
+++ b/buildkite.go
@@ -66,11 +66,12 @@ type Client struct {
 	httpDebug  bool
 }
 
-type clientOpt func(*Client) error
+// ClientOpt is a function that configures a Client.
+type ClientOpt func(*Client) error
 
 // WithHTTPClient configures the buildkite.Client to use the provided http.Client. This can be used to
 // customise the client's transport (e.g. to use a custom TLS configuration) or to provide a mock client
-func WithHTTPClient(client *http.Client) clientOpt {
+func WithHTTPClient(client *http.Client) ClientOpt {
 	return func(c *Client) error {
 		c.client = client
 		return nil
@@ -78,7 +79,7 @@ func WithHTTPClient(client *http.Client) clientOpt {
 }
 
 // WithBaseURL configures the buildkite.Client to use the provided base URL, instead of the default of https://api.buildkite.com/
-func WithBaseURL(baseURL string) clientOpt {
+func WithBaseURL(baseURL string) ClientOpt {
 	return func(c *Client) error {
 		var err error
 		c.BaseURL, err = url.Parse(baseURL)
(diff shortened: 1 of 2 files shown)


mcncl:
Thanks for the PR @prateek! On the face of it, this looks like a nice adjustment. We will review the PR and run tests against it before merging, assuming they pass.

mcncl (approved):
This looks fine to me and tests pass locally, we don't do a great job on testing third party contributions so if someone else wants to also take a look I'd welcome it, but I'm happy to merge this in.

@buildkite/support-engineering

mcncl:
@prateek sorry for the delay; we wanted to get a release out with a fix. It looks like linting isn't happy with this at the moment. I'd push up the fix but I can't given it's a fork. In your `example_test.go` you never use the `opts`, so we are getting an ineffectual assignment error, it would be good to use the opts and just not handle them;

```go
func Example_clientOptComposition() {
        // This is a test example only - we don't actually run the client
        // Simulating command-line flags or config values
        baseURL := "https://api.buildkite.com/"
        token := "your-token"
        userAgentFlag := "custom-agent" // Could be empty in some cases
        debugEnabled := false

        // Build options programmatically - we're testing the pattern more than actual execution
        var opts []buildkite.ClientOpt

        // Conditionally add other options
        if baseURL != buildkite.DefaultBaseURL {
                opts = append(opts, buildkite.WithBaseURL(baseURL))
        }

        if userAgentFlag != "" {
                opts = append(opts, buildkite.WithUserAgent(userAgentFlag))
        }

        if debugEnabled {
                opts = append(opts, buildkite.WithHTTPDebug(true))
        }

        // Always add required options
        opts = append(opts, buildkite.WithTokenAuth(token))

        // Create the client using the composed options
        client, err := buildkite.NewClient(opts...)
        if err != nil {
                // In a real application, you'd handle this error appropriately
                // For this example, we'll just acknowledge the client was created
                _ = client
        }
}
```

mcncl (approved):

mcncl:
@prateek all good, you submitted this how we'd want it to be done. We just need to better integrate third party submissions in CI, which we can definitely look in to

prateek:
Thanks!

just a thought: In the m3db/m3 repo I remember a mate had setup an input gate of some kind for 3p contributions (something like https://github.com/mvines/ci-gate iirc)

mcncl:
We do have that in places, and I thought we did here but I'll need to see. I think it's due to our clustering

mcncl:
@prateek just to keep you in the loop, we shipped https://github.com/buildkite/go-buildkite/pull/224 too to ensure we didn't need to do a new major release as this would have been a breaking change

prateek:
@mcncl you’re amazing mate, ta for the proactive comms!

Not trying to be a stickler, just wanna make sure I understand your point. How come this would be a breaking major? This diff is exporting something that wasn’t exported before, what could people have been doing before that this diff broke?

prateek:
lol, I totally get it - better safe than sorry :)

PR #221, opened 2025-05-23
Add lefthook

mcncl:
## Changes

- Adds a `lefthook.yml` file to the root of the project
    - Adds in some best pracise commands pre-commit

## Refs

- [lefthook](https://lefthook.dev/) can be installed locally and used to ensure _things_ are done at certain stages
    - In our `lefthook.yml` file, we ensure `go fmt` is run before the files are committed

### DIFF
--- a/lefthook.yml
+++ b/lefthook.yml
@@ -1,0 +1,8 @@
+pre-commit:
+  commands:
+    gofmt:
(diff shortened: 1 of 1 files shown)


PriyaSudip (approved):
Nice, good to know about lefthook!

PR #244, opened 2025-09-19
Setting min go version to 1.23 with toolchain version to 1.24

PriyaSudip:
PR to lower the minimum Go version to 1.23.

Ensures CI/dev still uses Go 1.24 via the toolchain.

### DIFF
--- a/go.mod
+++ b/go.mod
@@ -1,8 +1,8 @@
 module github.com/buildkite/go-buildkite/v4
 
-go 1.23.0
+go 1.23
 
-toolchain go1.24.0
+toolchain go1.24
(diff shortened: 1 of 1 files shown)


go.mod
```diff
@@ -1,6 +1,8 @@
 module github.com/buildkite/go-buildkite/v4
 
-go 1.24
+go 1.23.0
+
+toolchain go1.24.0
```
mcncl, line 5:
```suggestion
go 1.23

toolchain go1.24
```

mcncl (changes requested):

PR #247, opened 2025-09-22
Add changelog for v4.7.1

PriyaSudip:


mcncl (changes requested):
This has removed all other entries, that shouldn't be the case; the output of `ghch` can be inserted to the top of the file

PR #250, opened 2025-09-26
Fix URL resolution to preserve base URL path prefixes

bearcage-dayjob:
Replaces url.URL.ResolveReference() with custom resolveURL() method that properly handles base URLs containing path prefixes (e.g., https://proxy.example.com/buildkite/). The standard ResolveReference behaves differently when resolving relative vs absolute URLs, dropping the path prefix in one case and not the other, which leads to API calls to incorrect endpoints.

Changes:
- Add Client.resolveURL() method to handle relative/absolute urls w/ bases that have path prefixes
- Update Client.NewRequest() to use new URL resolution

Fixes issue where proxy/gateway deployments with path prefixes would lose the prefix during URL resolution, resulting in 404s.

### DIFF
--- a/buildkite.go
+++ b/buildkite.go
@@ -180,18 +180,54 @@ func (c *Client) populateDefaultServices() {
 	c.TestSuites = &TestSuitesService{c}
 }
 
+// resolveURL properly handles base URLs with path prefixes by combining the base URL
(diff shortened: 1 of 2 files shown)


nsuma8989 (approved):
Looks good to me

PR #251, opened 2025-10-19
fix: update cluster maintainers to match REST api for create and list

wolfeidau:
So at the moment I am getting the following error when I list clusters via the go-buildkite.

```
json: cannot unmarshal object into Go struct field Cluster.maintainers of type []buildkite.ClusterMaintainer
```

This problem was introduced in #240 which correctly specified the maintainers in the `ClusterCreate` however what we get back from the api differs from this. I have updated the tests to demonstrate the issue and added structures to support what is returned by the API.

### DIFF
--- a/clusters.go
+++ b/clusters.go
@@ -30,20 +30,20 @@ type ClusterUpdate struct {
 }
 
 type Cluster struct {
-	ID              string              `json:"id,omitempty"`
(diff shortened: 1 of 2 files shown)


CerealBoy (approved):
LGTM! 🚀 🚀

PR #306, opened 2026-05-20
feat: Make update payloads presence-aware

lox:
Update request structs currently use `omitempty`, so callers cannot distinguish "leave this field unchanged" from "send this field with an empty value". That makes valid PATCH operations silently no-op when callers try to clear fields such as pipeline tags, schedule env, descriptions, IP allowlists, or explicit boolean settings.

This PR takes the v5 break and makes update payloads presence-aware with `Optional[T]` and `Some(...)`. Unset optionals are omitted from the PATCH body, while `Some("")`, `Some(false)`, `Some([]string{})`, and `Some(map[string]string{})` are sent exactly as requested.

Examples:

```go
_, _, err := client.Pipelines.Update(ctx, org, pipelineSlug, buildkite.UpdatePipeline{
    Description:            buildkite.Some(""),
    Tags:                   buildkite.Some([]string{}),
    SkipQueuedBranchBuilds: buildkite.Some(false),
})
```

```go
_, _, err := client.PipelineSchedules.Update(ctx, org, pipelineSlug, scheduleID, buildkite.UpdatePipelineSchedule{
    Env:     buildkite.Some(map[string]string{}),
    Enabled: buildkite.Some(false),
})
```

The change also splits shared create/update inputs where create payloads should keep plain values and update payloads need presence tracking: cluster tokens, pipeline templates, teams, and test suites. The module path and examples move to `github.com/buildkite/go-buildkite/v5`, and the README/changelog include migration notes.

Validation:

- `mise exec go@1.25 -- go test ./...`
- `mise exec go@1.25 -- make test`

### DIFF
--- a/IMPLEMENTATION_NOTES.md
+++ b/IMPLEMENTATION_NOTES.md
@@ -4,7 +4,8 @@
 
 - Added `Optional[T]` and `Some[T]` as the presence primitive for PATCH request structs.
 - Confirmed the repo targets Go 1.25, so `json:",omitzero"` can use `Optional[T].IsZero()` to omit unset fields before `MarshalJSON` runs.
-- Chose to make explicit `null` count as present in `UnmarshalJSON`; absent and present-null should not collapse into the same state.
+- Kept `Optional[T]` request-only: it implements JSON marshaling and zero-value omission, but not unmarshaling.
+- Documented that `Some([]T(nil))` and `Some(map[K]V(nil))` encode as `null`; callers should use initialized empty collections when the API expects `[]` or `{}`.
 - Converted dedicated update structs to `Optional[T]`: clusters, cluster queues, cluster secrets, package registries, pipelines, pipeline schedules, team pipelines, and team suites.
 - Added raw request-body assertions for update tests so omitted keys and explicit empty values are visible to tests.
 - Added regression coverage for clearing `UpdatePipelineSchedule.Env`, clearing `UpdatePipeline.Tags`, and sending `skip_queued_branch_builds: false` deliberately.
--- a/IMPLEMENTATION_NOTES.md
+++ b/IMPLEMENTATION_NOTES.md
@@ -6,3 +6,6 @@
 - Added `Optional[T]` and `Some[T]` as the presence primitive for PATCH request structs.
 - Confirmed the repo targets Go 1.25, so `json:",omitzero"` can use `Optional[T].IsZero()` to omit unset fields before `MarshalJSON` runs.
 - Chose to make explicit `null` count as present in `UnmarshalJSON`; absent and present-null should not collapse into the same state.
+- Converted dedicated update structs to `Optional[T]`: clusters, cluster queues, cluster secrets, package registries, pipelines, pipeline schedules, team pipelines, and team suites.
+- Added raw request-body assertions for update tests so omitted keys and explicit empty values are visible to tests.
+- Added regression coverage for clearing `UpdatePipelineSchedule.Env`, clearing `UpdatePipeline.Tags`, and sending `skip_queued_branch_builds: false` deliberately.
--- a/IMPLEMENTATION_NOTES.md
+++ b/IMPLEMENTATION_NOTES.md
@@ -9,3 +9,6 @@
 - Converted dedicated update structs to `Optional[T]`: clusters, cluster queues, cluster secrets, package registries, pipelines, pipeline schedules, team pipelines, and team suites.
 - Added raw request-body assertions for update tests so omitted keys and explicit empty values are visible to tests.
 - Added regression coverage for clearing `UpdatePipelineSchedule.Env`, clearing `UpdatePipeline.Tags`, and sending `skip_queued_branch_builds: false` deliberately.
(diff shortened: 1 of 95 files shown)


lox:
@buildsworth-bk review please!

optional.go
```diff
@@ -0,0 +1,41 @@
...
+}
+
+// Value returns the wrapped value and whether it was set.
+func (o Optional[T]) Value() (T, bool) {
+	return o.value, o.set
+}
+
+// MarshalJSON encodes the wrapped value.
+func (o Optional[T]) MarshalJSON() ([]byte, error) {
+	return json.Marshal(o.value)
```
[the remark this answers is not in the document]
lox: Fixed in 6af9128: documented that `Optional[T]` must be used on fields tagged with `json:",omitzero"`, including the fallback behavior if the tag is missed.

IMPLEMENTATION_NOTES.md
```diff
@@ -0,0 +1,16 @@
...
+- Added raw request-body assertions for update tests so omitted keys and explicit empty values are visible to tests.
+- Added regression coverage for clearing `UpdatePipelineSchedule.Env`, clearing `UpdatePipeline.Tags`, and sending `skip_queued_branch_builds: false` deliberately.
+- Split shared create/update structs for cluster tokens, pipeline templates, teams, and test suites. Create structs keep plain values; update structs use `Optional[T]`.
+- Fixed `TestSuitesService.Update` to return the request construction error instead of `nil`.
+- Added raw request-body coverage for shared-type updates, including explicit false for pipeline template availability and team permission booleans.
+- Updated the module path and examples from `/v4` to `/v5` because the exported request structs are intentionally breaking.
+- Moved `buildkite-mcp-server` onto the local `/v5` module so its update tools preserve clear intent. The MCP branch keeps a temporary `replace github.com/buildkite/go-buildkite/v5 => ../go-buildkite` until v5 is tagged or otherwise consumable, and keeps a v4 client only for `buildkite-logs`, which has not migrated yet.
```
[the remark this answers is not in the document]
lox: Fixed in 6af9128: made the notes durable by removing the branch name and local MCP `replace` workaround, leaving only the SDK design/testing decisions that should not go stale.

lox:
@buildsworth-bk can you review again? I've addressed your concerns.

test_helpers_test.go
```diff
@@ -1,9 +1,47 @@
...
+	}
+}
+
+func assertRequestJSON(t *testing.T, r *http.Request, want string) {
+	t.Helper()
+
+	got, err := io.ReadAll(r.Body)
+	if err != nil {
+		t.Fatalf("reading request body: %v", err)
+	}
```
DrJosh9000, line 47: These assert helpers suffer from some of the drawbacks of asserts generally,  but the one that stands out to me is they `t.Fatalf` on JSON unmarshaling error (https://google.github.io/styleguide/go/decisions.html#keep-going). This is particularly a problem because `assertRequestJSON` is being called within the `httptest` server, which will be off the test's main goroutine (https://pkg.go.dev/testing#T.FailNow) 

It is what the existing tests do, so we can fix this later if you like, but I would turn these `t.Fatalf` calls into `t.Errorf` (probably with `return` afterwards, since the `cmp.Diff` is probably useless if the JSON didn't parse.)

optional_test.go
```diff
@@ -0,0 +1,106 @@
...
+		t.Errorf("Name = %q, want empty string", name)
+	}
+
+	tags, tagsOK := got.Tags.Value()
+	if !tagsOK {
+		t.Fatal("Tags was not marked set")
+	}
+	if len(tags) != 0 {
+		t.Errorf("Tags length = %d, want 0", len(tags))
+	}
```
DrJosh9000: A whole-structure `cmp.Diff` comparison might be a good choice here:

```suggestion
	want := payload{
		// Name should be unset
		Tags: Some([]string{}),
	}
	if diff := cmp.Diff(want, got); diff != "" {
		t.Errorf("Unmarshaled payload diff (-want +got):\n%s", diff)
	}
```

DrJosh9000 (approved):
Overall LGTM, I don't have any more comments for now.

test_helpers_test.go
lox: Fixed in 878f353: changed the request JSON helpers to use `t.Errorf` and return on read/parse failures, so they no longer call `FailNow` from httptest handler goroutines.

wolfeidau (approved):
👍🏻 🚀

test_suites_test.go
```diff
@@ -221,7 +216,7 @@ func TestTestSuitesService_Update(t *testing.T) {
 			}`)
 	})
 
-	got, _, err := client.TestSuites.Update(context.Background(), "my-great-org", suite.Slug, TestSuite{DefaultBranch: "default"})
+	got, _, err := client.TestSuites.Update(context.Background(), "my-great-org", suite.Slug, TestSuiteUpdate{DefaultBranch: Some("default")})
 	if err != nil {
 		t.Errorf("Pipelines.Update returned error: %v", err)
```
mcncl:
```suggestion
		t.Errorf("TestSuites.Update returned error: %v", err)
```

mcncl (approved):

PR #307, opened 2026-05-21
Fix: expose expires_at on AccessToken (SUP-7112)

stephanieatte:
This PR adds the `ExpiresAt` field to the `AccessToken` struct so that `expires_at` is correctly returned when calling the `/v2/access-token` API endpoint.

Previously the field was missing from the struct, causing the value to be silently dropped during JSON unmarshalling even though the API was returning it.

## Changes

- Added `expires_at` to the AccessToken struct in `access_tokens.go`
- Added `TestAccessTokensService_Get_WithExpiresAt` test to verify `expires_at` is correctly unmarshalled from the API response


access_tokens_test.go
```diff
@@ -57,3 +57,20 @@ func TestAccessTokensService_Revoke(t *testing.T) {
 		t.Errorf("AccessTokens.Revoke diff: (-got +want)\n%s", diff)
 	}
 }
+
+func TestAccessTokensService_Get_WithExpiresAt(t *testing.T) {
```
lizrabuya: I think it would be better if we keep all field assertions for `AccessToken` in one test function `TestAccessTokensService_Get`, rather than adding a separate test just to verify `expires_at`. Consolidating all field assertions in one test for a struct is cleaner and also reduces maintenance overhead.

lizrabuya (changes requested):

access_tokens_test.go
stephanieatte: Good call, I've consolidated the tests

lizrabuya (approved):

PR #309, opened 2026-05-26
ci: Add mise toolchain setup

lox:
Contributors currently have to line up their local Go, golangci-lint, and hook tooling by hand, while Buildkite pins those tools indirectly through Docker images. That makes local validation drift from CI easier than it needs to be.

This adds a repo-local `mise.toml` and generated `mise.lock` so contributors can run the same pinned toolchain with commands like `mise run lint`, `mise run test`, and `mise run ci`. The pins match the current repo target: Go 1.25.x, golangci-lint 2.8.0, and lefthook 2.1.6.

Buildkite now uses the setup-go plugin to install/export the Go toolchain and wire Go caches, while keeping `mise#v1.1.2` on the lint step for non-Go tools such as golangci-lint. The setup-go plugin does not have a version tag yet, so the pipeline pins the current plugin commit instead of floating on `main`.

The lint script also compares before/after diffs for tidy, format, and generate checks so an unrelated dirty worktree no longer gets reported as go.mod drift during local runs.


lox:
@buildsworth-bk review please

mcncl (approved):

PR #319, opened 2026-06-11
feat: improve rate limit retry handling in go-buildkite

Megh03:
- Retry all HTTP methods on 429 (not just GET)
  - Read RateLimit-Reset header for wait duration, fall back to Retry-After, then exponential backoff
  - Add WithRateLimitNotify callback so consumers can observe retries
  - Add WithMaxRetries to make retry limit configurable (default 3)
  - Replace cenkalti/backoff with github.com/buildkite/roko for retry loop management

### DIFF
--- a/buildkite.go
+++ b/buildkite.go
@@ -17,8 +17,8 @@ import (
 	"strings"
 	"time"
 
-	"github.com/buildkite/roko"
 	"github.com/buildkite/go-buildkite/v5/internal/bkmultipart"
+	"github.com/buildkite/roko"
 	"github.com/google/go-querystring/query"
 )
 
@@ -29,6 +29,10 @@ const (
 
 // DefaultUserAgent is the default user agent used for API requests
 var DefaultUserAgent = "go-buildkite/" + Version
+
+// errRateLimited is returned from the roko callback to signal a retryable 429.
+// It is never surfaced to callers — checkResponse produces the real *ErrorResponse.
+var errRateLimited = errors.New("rate limited")
 
 // Client - A Client manages communication with the buildkite API.
 type Client struct {
@@ -473,6 +477,9 @@ func (c *Client) Do(req *http.Request, v interface{}) (*Response, error) {
 		delay := retryDelay(resp, rt.AttemptCount())
 		rt.SetNextInterval(delay)
 
+		// roko calls MarkAttempt() after this callback returns, so AttemptCount()
+		// here is still the 0-based index of the current attempt. The last allowed
+		// attempt is index c.maxRetries (= WithMaxAttempts(c.maxRetries+1) - 1).
 		if rt.AttemptCount() < c.maxRetries {
 			// More retries remaining — drain and close so the connection can be reused.
 			_, _ = io.Copy(io.Discard, resp.Body)
@@ -486,12 +493,15 @@ func (c *Client) Do(req *http.Request, v interface{}) (*Response, error) {
 				fmt.Printf("DEBUG rate limited, retry %d in %v\n", rt.AttemptCount()+1, delay)
 			}
 		}
-		// On the final attempt, resp is kept intact so checkResponse can produce
-		// a proper *ErrorResponse from the 429 body.
+		// On the final attempt resp is kept intact so checkResponse can produce
+		// a proper *ErrorResponse from the 429 body rather than returning errRateLimited.
 
-		return errors.New("rate limited")
+		return errRateLimited
 	})
 
+	// resp is nil only when a network error or body-rewind error caused an early
+	// exit via rt.Break(). For exhausted 429s, resp is kept so checkResponse below
+	// can return a proper *ErrorResponse.
 	if resp == nil {
 		return nil, rokoErr
 	}
--- a/buildkite.go
+++ b/buildkite.go
@@ -5,6 +5,7 @@ import (
 	"bytes"
 	"context"
 	"encoding/json"
+	"errors"
 	"fmt"
 	"io"
 	"math/rand/v2"
@@ -16,6 +17,7 @@ import (
 	"strings"
 	"time"
 
+	"github.com/buildkite/roko"
 	"github.com/buildkite/go-buildkite/v5/internal/bkmultipart"
 	"github.com/google/go-querystring/query"
 )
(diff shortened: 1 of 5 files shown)


moskyb:
worth noting that we have a homegrown retry library [called roko](https://github.com/buildkite/roko) that does most of the things you've implemented from scratch here. no pressure at all to use it rather than this, but worth a callout. An example of it in use: https://github.com/buildkite/agent/blob/main/clicommand/agent_pause.go#L70

Megh03:
> worth noting that we have a homegrown retry library [called roko](https://github.com/buildkite/roko) that does most of the things you've implemented from scratch here. no pressure at all to use it rather than this, but worth a callout. An example of it in use: https://github.com/buildkite/agent/blob/main/clicommand/agent_pause.go#L70

Thanks @moskyb ! I wish I had known it sooner, would have saved a lot of effort 😅

mcncl:
@Megh03 I appreciate that it's additional effort, but would it be worth refactoring to include the `roko` library here so that it doesn't require such in the future?

buildkite.go
```diff
@@ -357,24 +407,77 @@ func (r *Response) populatePageValues() {
...
 }
 
+// retryDelay returns how long to wait before the next attempt after a 429.
+// It checks RateLimit-Reset first, then Retry-After, then falls back to capped
+// exponential backoff. Both headers are interpreted as delta-seconds (time until
+// reset) per the Buildkite API spec — not Unix timestamps.
+// A random jitter of up to 1 second is added on all paths to spread out
+// concurrent clients that receive the same reset time.
+func retryDelay(resp *http.Response, attempt int) time.Duration {
+	for _, header := range []string{"RateLimit-Reset", "Retry-After"} {
```
jeremybumsted: I don't think we send `Retry-After` as a response header, so this ends up being dead code afaik.

buildkite.go
```diff
@@ -383,33 +486,49 @@ func (c *Client) Do(req *http.Request, v interface{}) (*Response, error) {
...
 		}
 
-		respCh <- resp
-		return nil
-	}
+		delay := retryDelay(resp, rt.AttemptCount())
+		rt.SetNextInterval(delay)
 
-	notify := func(err error, delay time.Duration) {
+		if c.rateLimitNotify != nil {
```
jeremybumsted: I think this call is going to run on every 429 response, even if the limits have been exhausted (and thus no retry will be made).

I think that's fine if the intent is for the caller to have an accurate count of the number of 429s received, but for other use cases that are more interested in `status` == `429`, and the delay, the final delay is going to show as `60s` but the client is giving up.

Might be worth changing how we return that final attempt

buildkite.go
```diff
@@ -73,10 +79,18 @@ type Client struct {
 	TestRuns                 *TestRunsService
 	TestSuites               *TestSuitesService
 
-	authHeader string
-	httpDebug  bool
+	authHeader      string
+	httpDebug       bool
+	rateLimitNotify RateLimitNotify
+	maxRetries      int
```
jeremybumsted, line 85: The retry tests added in this PR pay ~13s of real `time.Sleep` during backoff (verified locally). Consider adding an unexported sleep seam here so the tests can stub the backoff:

```go
sleepFunc func(time.Duration) // test-only override for retry backoff; nil uses real time
```

When `nil` (i.e. all real usage), behaviour is unchanged.

buildkite.go
```diff
@@ -357,24 +407,77 @@ func (r *Response) populatePageValues() {
...
 func (c *Client) Do(req *http.Request, v interface{}) (*Response, error) {
-	respCh := make(chan *http.Response, 1)
+	var resp *http.Response
+
+	// roko requires a strategy, but the real delay is always driven by the
+	// server response (RateLimit-Reset header or exponential fallback), so
+	// Constant(0) is a required placeholder that is always overridden.
+	retrier := roko.NewRetrier(
+		roko.WithMaxAttempts(c.maxRetries+1),
+		roko.WithStrategy(roko.Constant(0)),
```
jeremybumsted: roko already exposes `WithSleepFunc`, so we can wire the seam in here to make the tests instant:

```go
retrierOpts := []roko.RetrierOpt{
    roko.WithMaxAttempts(c.maxRetries + 1),
    roko.WithStrategy(roko.Constant(0)),
}
if c.sleepFunc != nil {
    retrierOpts = append(retrierOpts, roko.WithSleepFunc(c.sleepFunc))
}
retrier := roko.NewRetrier(retrierOpts...)
```

roko's `sleepOrDone` still selects on `ctx.Done()` even with a stubbed sleep, so context cancellation keeps working.

retry_test.go
```diff
@@ -0,0 +1,490 @@
...
+	"time"
+)
+
+// TestDo_POST_IsRetriedOn429 verifies that POST requests are retried on 429.
+// The old Do() only retried GET requests; this confirms all methods are retried.
+// Intentionally omits RateLimit-Reset to exercise the exponential fallback path.
+func TestDo_POST_IsRetriedOn429(t *testing.T) {
+	callCount := 0
+
+	ms, client, teardown := newMockServerAndClient(t)
```
jeremybumsted: This suite takes ~13s because each retry does a real backoff sleep. With the `sleepFunc` seam on the client, we can add an instant-backoff helper and use it in the sleep-heavy retry tests:

```go
func newRetryTestClient(t *testing.T) (*mockServer, *Client, func()) {
    ms, client, teardown := newMockServerAndClient(t)
    client.sleepFunc = func(time.Duration) {} // skip real backoff sleeps
    return ms, client, teardown
}
```

Then swap `newMockServerAndClient(t)` → `newRetryTestClient(t)` in the ~8 sleep-heavy tests.

Two deliberate exceptions:
- `TestDo_ContextCancelledDuringRetryWait` should stay on the real sleep — its whole point is verifying a real wait is interrupted by context cancellation.
- `TestRetryDelay` needs no change; it calls `retryDelay` directly and never sleeps.

jeremybumsted (commented):
@Megh03 Nice work!

Couple of minor things that aren't really blocking, but I think would be good to tidy up before we merge.

Megh03:
Thanks @jeremybumsted for the review! All 5 addressed in the latest push 😬

jeremybumsted (approved):
Nice one @Megh03 !

PR #320, opened 2026-06-12
SUP-6268: Add Find test capability to TestsService

dahtey-bk:
As per https://linear.app/buildkite/issue/SUP-6268/add-list-and-find-to-testsservice , this PR is to add the capability to utilise the Find a Test REST API as per our docs: https://buildkite.com/docs/apis/rest-api/test-engine/tests#find-a-test-with-scope-and-name

Test case has been added and ran successfully.

### DIFF
--- a/tests.go
+++ b/tests.go
@@ -23,6 +23,11 @@ type Test struct {
 	FileName string `json:"file_name,omitempty"`
 }
 
+type FindTest struct {
+	Scope string `json:"scope"`
+	Name  string `json:"name"`
+}
+
 func (ts *TestsService) Get(ctx context.Context, org, slug, testID string) (Test, *Response, error) {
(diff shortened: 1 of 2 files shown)


tests.go
```diff
@@ -23,6 +23,11 @@ type Test struct {
 	FileName string `json:"file_name,omitempty"`
 }
 
+type FindTest struct {
+	Scope string `json:"scope"`
+	Name  string `json:"name"`
+}
```
mcncl: We might want to make this `FindTestOptions`, as per other _options_ structs

mcncl (changes requested):

tests.go
dahtey-bk: @mcncl sure thing, renamed it now :)

mcncl (approved):

PR #323, opened 2026-06-16
PB-1933 Add promised exit status fields to Job

123sarahj123:
### Description

Adds two fields to the `Job` struct so API consumers can read a running job's early failure declaration:

- `PromisedExitStatus *int` → `promised_exit_status`
- `PromisedExitStatusAt *Timestamp` → `promised_exit_status_at`

Both are `nil` until the job promises a final exit status. The Buildkite REST job payload already serialises these fields (via `API::JobPresenter`, emitted only when a promised status is present), so this is a purely additive struct change to expose data the API already returns.

Context: this is the first dependency for teaching `bk preflight` to surface early ("promised") job failures while a build is still running, ahead of the build cascading to failing.

### Changes

- `jobs.go`: add `PromisedExitStatus` and `PromisedExitStatusAt` to `Job`.
- `jobs_test.go`: add a decode test asserting the JSON tags populate the fields,
  and that absent fields stay `nil`.

### Testing

- `go test ./...` passes.
- `go fmt ./...` clean.

### Disclosures / Credits

Implemented with the assistance of Amp (AI). I reviewed and verified the change and tests locally.

### DIFF
--- a/jobs.go
+++ b/jobs.go
@@ -55,6 +55,14 @@ type Job struct {
 	TriggeredBuild     *TriggeredBuild `json:"triggered_build,omitempty"`
 	Priority           *JobPriority    `json:"priority"`
 	Step               *StepInfo       `json:"step,omitempty"`
+
+	// PromisedExitStatus is the exit status a running job has declared it will
+	// finish with, as part of an early failure declaration. Nil until the job
+	// promises a final exit status.
+	PromisedExitStatus *int `json:"promised_exit_status,omitempty"`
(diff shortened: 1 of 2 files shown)


123sarahj123:
@buildsworth-bk review this pr

mcncl (approved):

123sarahj123:
@mcncl Yay thank you so much for the quick review. 
What is the usual release cadence? I am working on some other stuff (not yet in review) that will be blocked on merge so wondering what that timeline looks like :)

PR #327, opened 2026-06-16
Add Jobs API list and get support

wolfeidau:


examples/jobs/get/main.go
```diff
@@ -0,0 +1,45 @@
...
+
+	"github.com/alecthomas/kingpin/v2"
+	"github.com/buildkite/go-buildkite/v5"
+)
+
+var (
+	apiToken = kingpin.Flag("token", "API token").Required().String()
+	baseURL  = kingpin.Flag("base-url", "Buildkite API base URL").Default(buildkite.DefaultBaseURL).String()
+	org      = kingpin.Flag("org", "Organization slug").Required().String()
+	pipeline = kingpin.Flag("pipeline", "Pipeline slug").Required().String()
```
mitchbne: org and pipeline can be optional

According to the docs (https://buildkite.com/docs/apis/rest-api/jobs#get-a-job), there's two ways to get a job:

curl -H "Authorization: Bearer $TOKEN" \
  -X GET "https://api.buildkite.com/v2/organizations/{org.slug}/pipelines/{pipeline.slug}/builds/{build.number}/jobs/{job.id}"

curl -H "Authorization: Bearer $TOKEN" \
  -X GET "https://api.buildkite.com/v2/organizations/{org.slug}/jobs/{job.id}"

Should we support this second way?

wolfeidau: @mitchbne good call, code was updated to add JobByOrg which just takes org slug and job id.

mcncl (approved):

PR #330, opened 2026-06-19
SUP-6270: Move ListEmojis from Client to an EmojisService

dahtey-bk:
As per https://linear.app/buildkite/issue/SUP-6270/move-listemojis-from-client-to-an-emojisservice , this PR is to create an EmojisService following similar pattern as other ones like Packages, etc.
Deprecated the misc.go file and client.ListEmojis since the creation of a dedicated emojis.go file containing EmojisService. _This could potentially be a breaking change so go-buildkite might need to be upgraded to v6._
Moved the Token struct to access_token.go in case it's still needed (wasn't sure to keep or remove as well)


misc.go
```diff
@@ -1,37 +0,0 @@
...
-// Emoji emoji, what else can you say?
-type Emoji struct {
-	Name string `json:"name,omitempty"`
-	URL  string `json:"url,omitempty"`
-}
-
-// ListEmojis list all the emojis for a given account, including custom emojis and aliases.
-//
-// buildkite API docs: https://buildkite.com/docs/api/emojis
-func (c *Client) ListEmojis(ctx context.Context, org string) ([]Emoji, *Response, error) {
```
omehegan, line 17: You're right in your PR description that this is a breaking change. By just deleting this, we'll break any existing users of the feature. I think in Go you can mark a function as `Deprecated` instead of deleting it. That will generate a warning so they can update their code. @mcncl is that what we should do here? I don't have the full context on how widely this feature might be used.

emojis.go
```diff
@@ -0,0 +1,39 @@
...
+//
+// buildkite API docs: https://buildkite.com/docs/api/emojis
+type EmojisService struct {
+	client *Client
+}
+
+type Emoji struct {
+	Name    string   `json:"name,omitempty"`
+	URL     string   `json:"url,omitempty"`
+	Aliases []string `json:"aliases,omitempty"`
```
omehegan, line 19: I notice that we're adding Aliases to the struct here. Why?

omehegan (changes requested):

misc.go
mcncl: Yeah, I think supporting both for now would be ideal, in a sense that we don't want to release v6 just for this.

Here is the documentation around marking code as deprecated: https://go.dev/doc/comment#deprecations

emojis.go
dahtey-bk: @omehegan  because as per our REST API, Aliases could be returned from the endpoint:
image

https://buildkite.com/docs/apis/rest-api/emojis

omehegan: I see, so it's something we could have been supporting in go-buildkite before, we just weren't. 👍

misc.go
dahtey-bk: Added the deprecated comment now @omehegan

omehegan (approved):

PR #335, opened 2026-07-05
feat: add missing documented fields to Agent struct

wolfeidau:
Add os_id, arch, queue, connected_at, disconnected_at, and lost_at, stopped_at fields to match the current Buildkite agents REST API docs (https://buildkite.com/docs/apis/rest-api/agents), which were previously missing from the Go struct.


mcncl (approved):
Tell you what, this library could do with an overhaul of the testing

Issue #349, opened 2026-07-20
Job.Signal is typed *int but the REST API returns a string


davidenglishmusic:
**Describe the bug**
`Job.Signal` is declared as `*int`, but the Buildkite REST API returns the job
`signal` field as a **string** (e.g. `"SIGKILL"`, `"SIGTERM"`). Any client that
decodes a build/job which has a non-null `signal` fails with:

`json: cannot unmarshal string into Go struct field Job.jobs.signal of type int`

**To Reproduce**
Decode any build whose job carries a signal (e.g. a cancelled/timed-out job).
Easiest via the Buildkite CLI, which vendors this library:
```
$ bk --version
bk version 3.54.0        # bundles go-buildkite v5.7.0

$ bk build list --pipeline <org>/<pipeline> --branch <branch> --limit 1 --json
Error: failed to list builds: json: cannot unmarshal string into Go struct field Job.jobs.signal of type int
```

**Expected behavior**
Job.Signal should unmarshal a string value returned by the REST API.

scadu:
Thanks for raising this, and for sending the PR over! 🙇 
We've just cut a new release with this change: https://github.com/buildkite/go-buildkite/releases/tag/v5.8.0

PR #350, opened 2026-07-20
Match Job.Signal type to that of API

davidenglishmusic:
The Job.Signal was declared as *int, but the Buildkite REST API returns the job
signal field as a string. This PR matches the type correctly.

https://github.com/buildkite/go-buildkite/issues/349


scadu (approved):
Thanks for that @davidenglishmusic!

PR #364, opened 2026-08-18
Add JobStateCounts to Build

nethsix:
### Description

The get-build API endpoint now returns a `job_state_counts` property (buildkite/buildkite#32263):

```json
"job_state_counts": { "total": 4, "states": { "passed": 2, "failed": 1, "running": 1 } }
```

It tallies every job in the build by state, regardless of any job filters applied to the request, so consumers can confirm a build's problem jobs are complete without a follow-up jobs listing. This PR decodes it into a new `Build.JobStateCounts` field (`*JobStateCounts`, nil when the API omits the property — it is only returned by builds#show).

First consumer: buildkite-mcp-server's `get_build_failure_summary` (PB-2566), which will propagate the counts once this ships.

### Changes

- `builds.go`: new `JobStateCounts` type (`total` + per-state `states` map); `Build.JobStateCounts *JobStateCounts` decoded from `job_state_counts`.
- `builds_test.go`: `TestBuildsService_Get` subtest decoding the property.

Additive only — no behavior change for existing fields.

### DIFF
--- a/builds.go
+++ b/builds.go
@@ -97,6 +97,13 @@ type TestEngineProperty struct {
 	Runs []TestEngineRun `json:"runs,omitempty"`
 }
 
+// JobStateCounts is a tally of a build's jobs grouped by state, as returned
+// by the get-build API endpoint.
+type JobStateCounts struct {
+	Total  int            `json:"total"`
+	States map[string]int `json:"states,omitempty"`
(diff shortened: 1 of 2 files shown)


wolfeidau (approved):
👍🏻

PR #366, opened 2026-08-19
Generate "Analytics" structs from OpenAPI spec

malclocke:
Adds a command to the Makefile to pull the Test Engine (aka "Analytics") OpenAPI spec and uses oapi-codegen to generate the related structs from the spec.


meghan-kradolfer (commented):
Are these struct changes going to break any existing consumers? Also, this is adding some internal APIs to a public space (for example `create_suite_from_pipeline`) are we ok with that?

malclocke:
> Are these struct changes going to break any existing consumers?

I don't think so. There was one point that was raised by the bot review, the addition of `suite.oidc_policy` to the response could be breaking. So this field is excluded from the `Suite` type.

malclocke:
Closing this PR for now, we may follow up again in the future.

PR #374, opened 2026-08-29
Add Teams to CreatePipeline

JuanitoFatas:
`CreatePipeline` only carries `team_uuids`. The [REST API docs](https://buildkite.com/docs/apis/rest-api/pipelines#create-a-yaml-pipeline) mark that parameter deprecated and replace it with `teams`, which maps each team UUID to an access level (`read_only`, `build_and_read`, `manage_build_and_read`). Callers cannot set an access level today.

Adds `Teams map[string]string`. `TeamUuids` stays, so existing callers still compile.

The deprecation note sits in the `CreatePipeline` doc comment, not on the field. A `// Deprecated:` marker on the field would be machine-readable, but any full-line comment inside a struct ends gofmt's field alignment group and splits the block. Prose in the struct doc keeps one aligned block, at the cost of the staticcheck/gopls warning.

`teams` also accepts a plain array of UUIDs, which takes the organization default access level. The map form is the one that needs the client change, so this adds only that.

### DIFF
--- a/pipelines.go
+++ b/pipelines.go
@@ -15,6 +15,10 @@ type PipelinesService struct {
 }
 
 // CreatePipeline - Create a Pipeline.
+//
+// Teams maps a team UUID to its access level on the pipeline: read_only,
+// build_and_read or manage_build_and_read. It supersedes TeamUuids, which is
+// deprecated because it cannot carry an access level.
 type CreatePipeline struct {
 	Name       string `json:"name"`
 	Repository string `json:"repository"`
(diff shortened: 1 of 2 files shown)


pipelines.go
```diff
@@ -36,9 +36,13 @@ type CreatePipeline struct {
...
 	CancelRunningBranchBuilds       bool              `json:"cancel_running_branch_builds"`
 	CancelRunningBranchBuildsFilter string            `json:"cancel_running_branch_builds_filter,omitempty"`
-	TeamUuids                       []string          `json:"team_uuids,omitempty"`
-	Visibility                      string            `json:"visibility,omitempty"`
-	Tags                            []string          `json:"tags,omitempty"`
+	// Deprecated: use Teams, which also carries each team's access level.
+	TeamUuids []string `json:"team_uuids,omitempty"`
+	// Teams maps a team UUID to its access level on the pipeline: read_only,
+	// build_and_read or manage_build_and_read.
+	Teams      map[string]string `json:"teams,omitempty"`
```
nethsix: Maybe the formatting?

```suggestion
	Teams                       map[string]string `json:"teams,omitempty"`
```

nethsix (approved):
Maybe just indentation? Otherwise looks good.

JuanitoFatas:
Thanks @nethsix 🙇🏻

PR #375, opened 2026-08-30
Add TestSuitesService.ListByPipeline

malclocke:
Adds support for the `listPipelineSuites` operation from the [Test Engine OpenAPI spec](https://api.buildkite.com/v2/analytics/openapi.yaml).

```go
suites, _, err := client.TestSuites.ListByPipeline(ctx, org, pipelineSlug, nil)
```

`GET /v2/analytics/organizations/{org}/pipelines/{pipeline_slug}/suites` returns the suites that have recorded at least one run attributed to the pipeline. It responds with the same `Suite` schema as `listSuites`/`getSuite`, so the new method decodes into the existing `TestSuite` type and takes `TestSuiteListOptions` for pagination — mirroring how `BuildsService.ListByPipeline` reuses `BuildsListOptions`.

Also adds an `examples/test_suites/list_by_pipeline` example alongside the existing `list` one.

Two things intentionally left out:

- The `Suite` schema in the spec also carries `organization_id`, `application_name`, `color`, `emoji`, `oidc_policy`, and `api_token`, none of which are on `TestSuite` today. Adding them would change `List`/`Get` too, so it felt like a separate change.
- The spec documents no pagination parameters for this operation, but the options struct is wired up for consistency with every other list method.

Closes [TE-6802](https://linear.app/buildkite/issue/TE-6802/update-go-buildkite-with-pipelinesuites-endpoint)

### DIFF
--- a/examples/test_suites/list_by_pipeline/main.go
+++ b/examples/test_suites/list_by_pipeline/main.go
@@ -1,0 +1,39 @@
+package main
+
+import (
+	"context"
+	"encoding/json"
+	"fmt"
+	"log"
+	"os"
+
+	"github.com/alecthomas/kingpin/v2"
+	"github.com/buildkite/go-buildkite/v5"
+)
+
+var (
+	apiToken = kingpin.Flag("token", "API token").Required().String()
+	org      = kingpin.Flag("org", "Organization slug").Required().String()
+	pipeline = kingpin.Flag("pipeline", "Pipeline slug").Required().String()
+)
+
+func main() {
+	kingpin.Parse()
+
+	client, err := buildkite.NewOpts(buildkite.WithTokenAuth(*apiToken))
+	if err != nil {
+		log.Fatalf("creating buildkite API client failed: %v", err)
+	}
+
+	suites, _, err := client.TestSuites.ListByPipeline(context.Background(), *org, *pipeline, nil)
+	if err != nil {
+		log.Fatalf("listing test suites for pipeline failed: %s", err)
+	}
+
+	data, err := json.MarshalIndent(suites, "", "\t")
+	if err != nil {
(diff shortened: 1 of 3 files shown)


test_suites.go
```diff
@@ -61,6 +61,30 @@ func (tss *TestSuitesService) List(ctx context.Context, org string, opt *TestSui
 	return testSuites, resp, err
 }
 
+// ListByPipeline returns the test suites that have recorded at least one run
+// attributed to the given pipeline, ordered by creation time. Each suite is
+// returned once.
+func (tss *TestSuitesService) ListByPipeline(ctx context.Context, org, pipelineSlug string, opt *TestSuiteListOptions) ([]TestSuite, *Response, error) {
```
malclocke, line 67: `TestSuiteListOptions` is composed of only `ListOptions`, which [contains](https://github.com/buildkite/go-buildkite/blob/2024b73133acd8b5d13a6261945f53ca461b435e/buildkite.go#L622-L628) `Page` and `PerPage` only. Both are supported by the endpoint.

examples/test_suites/list_by_pipeline/main.go
```diff
@@ -0,0 +1,39 @@
...
+	"github.com/buildkite/go-buildkite/v5"
+)
+
+var (
+	apiToken = kingpin.Flag("token", "API token").Required().String()
+	org      = kingpin.Flag("org", "Organization slug").Required().String()
+	pipeline = kingpin.Flag("pipeline", "Pipeline slug").Required().String()
+)
+
+func main() {
```
malclocke, line 20: Successfully tested this example against production.

meghan-kradolfer (approved):

PR #378, opened 2026-09-07
Add ExecutionsService.GetTrace

malclocke:
Implements the `getExecutionTrace` operation from the Test Engine (Analytics) REST API — the OpenTelemetry trace recorded for a single test execution, rolled up by category and ranked by the spans that spent the time.

```go
trace, _, err := client.Executions.GetTrace(ctx, org, suiteSlug, executionID, &buildkite.ExecutionTraceOptions{
    View: buildkite.TraceViewFull,
})
```

### What's here

- `ExecutionsService` with `GetTrace`, on `client.Executions`, hitting `v2/analytics/organizations/{org}/suites/{slug}/executions/{id}/trace`.
- `ExecutionTrace`, `TraceSpan` and `TraceCategoryRollup`, plus typed enums `TraceCategory` (`sql`/`db`/`cache`/`http`/`rpc`/`messaging`/`test`/`other`) and `TraceView` (`summary`/`full`), following the `RetryAgentAffinity` precedent in `cluster_queues.go`.
- An example under `examples/executions/trace`.
- Tests covering the full view, the summary view (no query string sent, `Spans` nil), an execution with no trace, and a 404.

### Naming decisions worth a look

- `*_ns` fields are spelled out — `DurationNanoseconds`, `SelfTimeNanoseconds`, `TraceDurationNanoseconds` — consistent with `duration_avg` becoming `DurationAverage` in `tests.go`.
- Nullable strings (`trace_id`, `parent_span_id`, `kind`, `service_name`, `scope_name`, `detail`, `status_message`) are plain `string` with `omitempty`, per repo convention. Nullable `duration` is `*float64`, matching `Reliability`.
- `Spans` is a plain slice, which preserves the null-vs-`[]` distinction the spec calls out: nil in the `summary` view, empty for a `full` view of a trace with no spans.

Closes TE-6922

🤖 Generated with [Claude Code](https://claude.com/claude-code)

PR #379, opened 2026-09-08
feat: add ExecutionsService.ListSlowestByBuild

mttns:
Adds `client.Executions.ListSlowestByBuild`, which answers "which tests took the longest in this build?" It returns the slowest executions first, across every suite the caller can see. Each result has the `SuiteSlug` and `ID` that `GetTrace` needs, so finding a slow test and seeing where its time went is two calls:

```go
executions, _, err := client.Executions.ListSlowestByBuild(ctx, org, buildUUID, &buildkite.SlowestExecutionsOptions{Limit: 5})
if err == nil && len(executions) > 0 && executions[0].HasTrace {
    trace, _, err := client.Executions.GetTrace(ctx, org, executions[0].SuiteSlug, executions[0].ID, nil)
    // Handle err and use trace.
}
```

`HasTrace` reflects the endpoint's `has_trace` hint added in [buildkite/buildkite#33970](https://github.com/buildkite/buildkite/pull/33970): a trace was recorded and is still within span retention. Check it before fetching a trace; even when true, the trace may be empty if spans were sampled away. The `--with-trace` example skips the request when the slowest execution has no trace.

The response is a top-N list, not a paginated collection, so the method returns a plain slice. `Limit` defaults to the server's 20 when left zero.

Includes tests, and an example under `examples/executions/slowest_by_build`.

Stacked on #378, which adds `ExecutionsService`. Closes TE-6981.

### DIFF
--- a/examples/executions/slowest_by_build/main.go
+++ b/examples/executions/slowest_by_build/main.go
@@ -1,0 +1,65 @@
+package main
+
+import (
+	"context"
+	"encoding/json"
+	"fmt"
+	"log"
+	"os"
+
+	"github.com/alecthomas/kingpin/v2"
+	"github.com/buildkite/go-buildkite/v5"
+)
+
+var (
+	apiToken  = kingpin.Flag("token", "API token").Required().String()
+	baseURL   = kingpin.Flag("base-url", "Buildkite API base URL").Default(buildkite.DefaultBaseURL).String()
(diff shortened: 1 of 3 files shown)


mttns:
Need to add `has_trace` to the response so that agents know not to send an extra request to retrieve a non-existent trace

mttns:
@buildsworth-bk review approve L2

PR #381, opened 2026-09-15
PB-3329: Add Job.BrokenReason from the REST API broken_reason field

nethsix:
### Description

The REST API returns `broken_reason` on every job — nil unless the job is `broken`, otherwise a reason code (`conditional_failed`, `parallelism_zero`, …) or the pipeline's free-text `skip` reason. `Job` had no field for it, so the value was dropped on decode. This adds a `BrokenReason` string field (JSON key `broken_reason`, omitempty) next to `SignalReason`, matching the API's field order, plus a decode test for a broken job.

### Context

- https://linear.app/buildkite/issue/PB-3329 — the consumer is buildkite-mcp-server, which currently cannot tell a config-skipped job from any other job that never ran

### Changes

- `jobs.go`: one field on `Job`
- `jobs_test.go`: `TestJobsService_ListByBuild_BrokenJob`

### Verification

`go test -timeout=3s ./...` and `golangci-lint run ./...` pass with the repo's pinned Go 1.26.6 / golangci-lint 2.12.2.

### Deployment

Additive and backwards-compatible. Needs a tagged release (v5.17.0) so buildkite-mcp-server can drop its temporary pseudo-version pin.

### Rollback

Safe to revert; no consumer depends on the field until the MCP server PR merges.


wolfeidau (approved):
👍🏻

PR #382, opened 2026-09-15
Report Analytics API response drift in CI + fix existing drift

malclocke:
## Why

Analytics response models can fall behind the published OpenAPI schema without any signal, leaving newly documented response fields unavailable to users of this client.

## What

Add a dedicated CI check that compares the JSON fields in supported Analytics response types with the live OpenAPI schemas. The step soft-fails and publishes a warning annotation containing the differences, so upstream schema changes are visible without blocking unrelated work.

Bring the currently drifting Suite, Test, and Execution response types into alignment with the published schema, with one commit per response type.


.buildkite/steps/analytics-openapi-drift.sh
```diff
@@ -0,0 +1,20 @@
+#!/usr/bin/env bash
+
+set -Eeuo pipefail
+
+if output=$(go run ./.buildkite/steps/analytics-openapi-drift 2>&1); then
+  exit 0
+fi
+
+printf '%s\n' "$output"
```
malclocke, line 19: Example soft failing annotation in https://buildkite.com/buildkite/go-buildkite/builds/1491/list

.buildkite/steps/analytics-openapi-drift/main.go
ashmckenzie, line 1: question: Should we add tests here?

.buildkite/steps/analytics-openapi-drift/main.go
```diff
@@ -0,0 +1,188 @@
...
+type schema struct {
+	properties map[string]bool
+	refs       []string
+}
+
+func main() {
+	schemas, err := fetchSchemas()
+	if err != nil {
+		fmt.Fprintln(os.Stderr, err)
+		os.Exit(1)
```
ashmckenzie: question: Should we use different exit codes for unexpected errors vs. when there's drift? e.g.


```suggestion
		os.Exit(2)
```

.buildkite/steps/analytics-openapi-drift/main.go
malclocke: I did initially add this check as a test, skipped by default, then changed it to a standalone `main.go` file. So functionally it kinda is a test, and testing the test feels a bit OTT - the check is advisory only anyway.

Happy to add it if you think it's warranted though.

ashmckenzie: It's fairly straightforward, let's omit the test for now.

ashmckenzie (approved):

